PR #6, opened 2017-11-20
initiating frontend dev

tjingrant:
There are several questions I'm currently still searching for answers to..
1. Best way to specify inputs/outputs. 
- Currently I take Placeholder node as an input.
- Currently I require the proto definition of output node to be passed in. 
2. Shape of output is not present in TF GraphDef.

I'll look into it more on Monday.

### DIFF
--- a/onnx_tf/frontend.py
+++ b/onnx_tf/frontend.py
@@ -1,0 +1,83 @@
+"""Frontend for exporting Tensorflow graph to ONNX graph
+
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import re
+
+from onnx import onnx_pb2, helper
+from onnx.onnx_pb2 import GraphProto, TensorProto, AttributeProto
+
+class TensorflowFrontend(object):
+  """ Tensorflow Frontend for ONNX
+  """
+
+  # refer to
+  # https://github.com/tensorflow/tensorflow/blob/f284c708a8f3e2672d41dd3e7f6f03b9d26f0c80/tensorflow/core/framework/types.proto
+
+  onnx_types = [
+  "invalid",
+  TensorProto.FLOAT,
+  TensorProto.DOUBLE,
+  # DT_INT32 = 3;
+  # DT_UINT8 = 4;
+  # DT_INT16 = 5;
+  # DT_INT8 = 6;
+  # DT_STRING = 7;
+  # DT_COMPLEX64 = 8;  // Single-precision complex
+  # DT_INT64 = 9;
+  # DT_BOOL = 10;
+  # DT_QINT8 = 11;     // Quantized int8
+  # DT_QUINT8 = 12;    // Quantized uint8
+  # DT_QINT32 = 13;    // Quantized int32
+  # DT_BFLOAT16 = 14;  // Float32 truncated to 16 bits.  Only for cast ops.
+  # DT_QINT16 = 15;    // Quantized int16
+  # DT_QUINT16 = 16;   // Quantized uint16
+  # DT_UINT16 = 17;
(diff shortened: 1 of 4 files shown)


onnx_tf/frontend.py
```diff
@@ -0,0 +1,83 @@
...
+  # DT_HALF = 19;
+  # DT_RESOURCE = 20;
+  # DT_VARIANT = 21;  // Arbitrary C++ data types
+  # DT_UINT32 = 22;
+  # DT_UINT64 = 23;
+  ]
+
+  @classmethod
+  def op_name_to_lower(cls, name):
+    return re.sub('(?<!^)(?=[A-Z])', '_', name).lower()
```
arpith-jacob: Is this also sanitizing names?  Could you add a comment on the rules?

test/frontend/test_node.py
```diff
@@ -0,0 +1,42 @@
...
+    for node in tf_graph.node:
+        if node.name == "Relu":
+            output_node = node
+
+    onnx_graph = convert_graph(tf_graph, output_node)
+    tf_rep = prepare(helper.make_model(onnx_graph))
+    in_tensor = self._get_rnd(shape)
+    output = tf_rep.run({"Placeholder": in_tensor})["Relu"]
+
+    np.testing.assert_allclose(output, np.clip(in_tensor, 0, 1.0))
```
arpith-jacob: Is it possible to compare the result of evaluating 'tf_graph' to 'output'?

onnx_tf/frontend.py
```diff
@@ -0,0 +1,83 @@
...
+      else:
+        handler_name = "handle_" + cls.op_name_to_lower(node.op)
+
+        # Check if specialized handler exists.
+        if handler_name in dir(cls):
+          method_to_call = getattr(cls, handler_name)
+          ops_proto.append(method_to_call(node))
+
+    return helper.make_graph(ops_proto,
+                            "onnx_graph",
```
arpith-jacob: Can you pass the name as an argument to this function?

onnx_tf/frontend.py
```diff
@@ -0,0 +1,83 @@
...
+  # DT_UINT64 = 23;
+  ]
+
+  @classmethod
+  def op_name_to_lower(cls, name):
+    return re.sub('(?<!^)(?=[A-Z])', '_', name).lower()
+
+  @classmethod
+  def tensorflow_graph_to_onnx_graph(cls, graph_def, output):
+    inputs_proto = []
```
arpith-jacob: It's not clear to me what 'output' and 'inputs_proto' mean.

onnx_tf/frontend.py
```diff
@@ -0,0 +1,83 @@
...
+        if handler_name in dir(cls):
+          method_to_call = getattr(cls, handler_name)
+          ops_proto.append(method_to_call(node))
+
+    return helper.make_graph(ops_proto,
+                            "onnx_graph",
+                            inputs_proto,
+                            [helper.make_tensor_value_info(output.name,
+                                                          cls.onnx_types[output.attr["T"].type],
+                                                          [10])])
```
arpith-jacob: Why the magic number [10]?

tjingrant:
I'm working on reversing the trivial maps from frontend.
After that we can distribute work as before.

test/frontend/test_node.py
```diff
@@ -0,0 +1,46 @@
...
+                    .astype(np.float32)
+
+  def test_relu(self):
+    shape = (10, 10)
+    x = tf.placeholder(tf.float32, shape=shape)
+    y = tf.nn.relu(x)
+
+    tf_graph = y.graph.as_graph_def(add_shapes=True)
+    for node in tf_graph.node:
+        if node.name == "Relu":
```
doru1004: Can we make this more robust or find a way of getting the output info directly from y?

PR #7, opened 2017-11-20
ONNX has changed Pad operator attribute from 'paddings' to 'pads'

bddppq:
https://github.com/onnx/onnx/pull/289


tjingrant:
Thanks!

PR #8, opened 2017-11-21
switch to python object instead of protobuf

tjingrant:


tjingrant:
I'm moving away from directly using protobuf object in the frontend translation because you can't edit attributes in place which is a great PITA.

onnx_tf/common.py
```diff
@@ -14,7 +15,7 @@
 def invertible(dict):
     # invertible iff one-to-one and onto
     # onto is guaranteed, so check one-to-one
-    return not (len(dict.values()) != len(dict.values()))
+    return not (len(dict.values()) != len(set(dict.values())))
```
arpith-jacob: Can you change the double negative to just: len(..) == len(set())?

arpith-jacob:
LG

PR #12, opened 2017-11-22
process constant inputs more reasonably

tjingrant:


tjingrant:
So I was working of translating tf.pad, but realized that tensorflow takes a tensor to specify pads whereas ONNX uses attribute "pad" to specify pads. The difference is that the value of a tensor may not be known at conversion time unless it is a constant tensor. Since ONNX requires pads to be an attribute, naturally we would restrict the padding tensor to be a constant in Tensorflow and introduce some mechanism to extract these constants at conversion time to be fed into the ONNX "pad" attribute. 

This patch does exactly that. For the same reason, unit tests are updated as well to handle constants.

tjingrant:
Merging due to https://github.com/onnx/onnx/pull/311 . Our master branch will otherwise keep failing CI.

PR #14, opened 2017-11-25
include license file in distributions

dougalsutherland:
Redistributions of your source code are supposed to include the `LICENSE` text under the terms of your license, but your distributions currently don't – this makes it harder for everyone to follow the license. :)

The `MANIFEST.in` file makes sure that `LICENSE` ends up in the `.tar.gz` source files from `sdist`, and the `setup.cfg` setting tells wheels about it.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include LICENSE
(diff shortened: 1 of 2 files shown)


tjingrant:
@dougalsutherland Thanks!

PR #34, opened 2017-12-18
Add concat handler in front-end

LucasMahieu:
*  Bug fix : give the real shape for Const node, not the flatten shape 
*  Add concat handler and test of concat in front-end 
*  Improve font end test_node


tjingrant:
Thanks, looks good!

PR #35, opened 2017-12-21
Add the flag data format and previous node to handler

LucasMahieu:
* Add a parameter for all handler front-end functions to give the previous nodes to the current node in the graph in order to be able to get the input shape tensor for examples
* Add a flag that user can specify in order to know when converting to/from TF which data format it is used. Then, it is possible to add transpose of to change parameters (axis, perm, shape, etc)

### DIFF
--- a/onnx_tf/backend.py
+++ b/onnx_tf/backend.py
@@ -132,26 +132,16 @@ class TensorflowBackend(Backend):
       "dot": tf.contrib.keras.backend.dot,
       "exp": tf.exp,
       "floor": tf.floor,
-      "gather": tf.gather,
       "log": tf.log,
       "neg": tf.negative,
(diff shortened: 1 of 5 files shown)


tjingrant:
I'm still reviewing most of it, but a quick thing to change is that we don't call a GPU device "GPU" in onnx. It is instrumental to differentiate between CUDA and OpenCL enabled device. Thus could you please change "GPU" to "CUDA". ONNX spec also refers to GPU as "CUDA", e.g.:
https://github.com/onnx/onnx/blob/9bc425d09a47ee75080c97e4a96887de97eac6ec/onnx/backend/base.py#L15

LucasMahieu:
I made an other pull request were I renamed GPU in CUDA.

PR #36, opened 2017-12-21
Add pooling ops

LucasMahieu:
* Add front-end support of avg_pool and max_pool


LucasMahieu:
Do you know why this error occurs just with python 3 and not with python 2 ?

LucasMahieu:
I will make proper pull request with pooling ops when PR #39 will be merged

PR #37, opened 2017-12-21
Change tf pad algo

LucasMahieu:
* Rewrite the tf algorithm computation according to [tf doc](https://www.tensorflow.org/api_guides/python/nn#Convolution)


LucasMahieu:
I will make proper pull request for tf padding algo when PR #39 will be merged

Issue #31, opened 2017-12-15
Force NCHW format


lukemurray:
Hi there, this seems to add a bunch of transpose ops into the graph, can we force it to build the graph in NCHW format?

tjingrant:
Are you building your graph on a CPU? Transpositions only occur when NHWC format is needed for computation (typically when you have no GPU). Fundamentally if you only have CPUs, there's no way you can execute that graph because TF does not have full support for NCHW format on CPU. (Although you may still want to build the graph that way despite the fact that you cannot run it, but I suspect this is not your concern.)

lukemurray:
I guess I want a typical tf model format actually. It is going into another third-party tool that doesn't currently support the transposes. I'm changing it so it transposes the raw data and therefore creates a cleaner tf graph without any transforms. Which this third-party tool will support.

Any reason you didn't implement it that way? I guess in general transpose is not expensive.

________________________________

tjingrant:
This is because we assume that the user will run the graph in the same machine as he/she converts the graph, which is the case for us, but may not be the case for others. We may support exporting GPU-compatible models on a CPU-only machine in the future. I'll take note and see what I can do.

lukemurray:
Yep, that makes sense. I think I can get this working for my case. But I would say the changes are nice enough for a PR 😁

________________________________

lukemurray:
Actually if I understand you correctly. I can load the onnx model and save the tf model on a computer with cuda and there will be no transforms in the tf graph?

________________________________

tjingrant:
Yes, we basically ask TF whether there's a CUDA device and if there's not, we perform transpositions. The condition is checked here:
https://github.com/onnx/onnx-tensorflow/blob/ca5c120d12282f11152ebc3ef29141f369953f66/onnx_tf/backend.py#L990

And the decision to prepend/append transpositions to an op is made in every op that concerns data format, for instance, the pooling op has the following logic:
https://github.com/onnx/onnx-tensorflow/blob/ca5c120d12282f11152ebc3ef29141f369953f66/onnx_tf/backend.py#L543

And contributions are always welcome!

LucasMahieu:
I have though a lot about this problem. Here is a proposal of how I think things have to be done to be optimal.

I will give an example to explain my idea:
I suggest that the user have the choice of two things:

    - The device on which he wants to run the TF graph
          (the source graph for front-end or the destination graph for back-end) (GPU or CPU)
[...]

PR #39, opened 2018-01-03
Add flag data format

LucasMahieu:
* Add a parameter for all handler front-end functions to give the previous nodes to the current node in the graph in order to be able to get the input shape tensor for examples
* Add a flag that user can specify in order to know when converting to/from TF which data format it is used. Then, it is possible to add transpose of to change parameters (axis, perm, shape, etc)
* Add a flag that user can specify in order to know when converting to/from TF for which device type the graph should be compatible : CPU or CUDA.

### DIFF
--- a/onnx_tf/frontend.py
+++ b/onnx_tf/frontend.py
@@ -25,6 +25,13 @@ from onnx.helper import (
 from onnx.onnx_pb2 import GraphProto, TensorProto, AttributeProto
 from tensorflow.python.framework.tensor_util import MakeNdarray
 
+def get_nodes_by_name(nodes, names):
+  ret = []
+  for node in nodes:
+    if node.name in names:
+      ret.append(TensorflowNode(node))
+  return ret
+
 class TensorflowNode(object):
 
   # Keyed by old attribute names.
@@ -152,10 +159,11 @@ class TensorflowFrontend(object):
 
         # Check if specialized handler exists.
         if handler_name in dir(cls):
+          input_node = get_nodes_by_name(graph_def.node, node.inputs)
           method_to_call = getattr(cls, handler_name)
-          ops_proto.append(method_to_call(node, consts))
+          ops_proto.append(method_to_call(node, consts, input_node))
         else:
-          raise NotImplementedError("{} op is not implemented.".format(node.op))
+          raise NotImplementedError("{} op is not implemented, handler {} do not exists.".format(node.op, handler_name))
 
     output = TensorflowNode(output)
     # making output proto
@@ -179,15 +187,15 @@ class TensorflowFrontend(object):
             onnx_op, node.inputs, [node.name], name=node.name, broadcast=1)
 
   @classmethod
-  def handle_logical_and(cls, node, consts):
+  def handle_logical_and(cls, node, consts, input_node):
     return cls._bin_op(node, "And")
 
   @classmethod
-  def handle_logical_or(cls, node, consts):
+  def handle_logical_or(cls, node, consts, input_node):
     return cls._bin_op(node, "Or")
 
   @classmethod
-  def handle_pad(cls, node, consts):
+  def handle_pad(cls, node, consts, input_node):
     assert node.inputs[1] in consts.keys()
     supported_modes = ["constant", "reflect"]
     mode = node.attr.get("mode", "constant")
@@ -204,7 +212,7 @@ class TensorflowFrontend(object):
             value=0.0)
 
   @classmethod
-  def handle_random_standard_normal(cls, node, consts):
+  def handle_random_standard_normal(cls, node, consts, input_node):
     """ Tensorflow does not have a generic random_normal op.
         The generic random_normal op is translated into a scaled
         and offsetted random standard normal op.
@@ -220,7 +228,7 @@ class TensorflowFrontend(object):
             shape=node.attr["_output_shapes"][0])
 
   @classmethod
-  def handle_random_uniform(cls, node, consts):
+  def handle_random_uniform(cls, node, consts, input_node):
     """ Tensorflow does not have a generic random_uniform op.
         The generic random_uniform op is translated into a scaled
         and offsetted random standard uniform op.
@@ -236,7 +244,7 @@ class TensorflowFrontend(object):
             shape=node.attr["_output_shapes"][0])
 
   @classmethod
-  def _reduce_op(cls, op, node, consts):
+  def _reduce_op(cls, op, node, consts, input_node):
     assert node.inputs[1] in consts.keys()
     axes = consts[node.inputs[1]]
     return helper.make_node(op,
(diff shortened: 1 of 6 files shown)


test/backend/test_model.py
```diff
@@ -27,6 +27,7 @@ def test_relu_node_inplace(self):
       inputs=[helper.make_tensor_value_info("X", TensorProto.FLOAT, [3, 2])],
       outputs=[helper.make_tensor_value_info("X1", TensorProto.FLOAT, [3, 2])])
     tf_rep = prepare(helper.make_model(graph_def))
+
```
tjingrant: These changes seem irrelevant, would you mind cleaning up a bit so the commit history will not be unnecessarily long?

test/frontend/test_node.py
```diff
@@ -38,13 +38,15 @@ class TestNode(unittest.TestCase):
...
-  test_option = test_data[5] if len(test_data) > 5 else {}
+  test_option = test_data[7] if len(test_data) > 7 else {}
 
   def do_test_expected(self):
     tf_op = test_data[1]
     output_name = test_data[2]
     inputs = test_data[3]
     attrs = test_data[4]
+    device = test_data[5]
+    channel_last = test_data[6]
```
tjingrant: Rather than specifying these two attributes, can we enumerate through all possibilities? For instance, if there's a GPU, then test NHWC and NCHW, if not, only test NHWC. Checks can be done via `supports_device` function.

This won't have any effects on CI, but I can test them locally on my work machine with GPU. This could significantly increase confidence in correctness.

onnx_tf/frontend.py
```diff
@@ -217,14 +264,23 @@ def handle_random_standard_normal(cls, node, consts):
...
         and offsetted random standard uniform op.
     """
+    shape = node.attr["_output_shapes"][0]
+    # Arrange shape to the according data format 
+    if cls._channel_last:
+      rank = len(shape)
+      if rank >= 3:
+        old_shape = shape[:]
+        for i, s in enumerate(shape):
+          j = i % (rank-1) + 1 if i != 0 else 0
```
tjingrant: I have seen this pattern quite a bit. Can you use a helper function to perform this NHWC->NCHW transformation? This could increase readability quite a bit.

onnx_tf/frontend.py
```diff
@@ -152,20 +168,25 @@ def tensorflow_graph_to_onnx_graph(cls, graph_def, output, name="graph"):
 
         # Check if specialized handler exists.
         if handler_name in dir(cls):
+          input_node = get_nodes_by_name(graph_def.node, node.inputs)
           method_to_call = getattr(cls, handler_name)
-          ops_proto.append(method_to_call(node, consts))
+          ops_proto.append(method_to_call(node, consts, input_node))
```
tjingrant: The whole point of having a input_node param seems to be that you want to access the shapes of input nodes, which might be better suited as an attribute in the TensorflowNode class? What do you think? This way we can keep the signature of handler clean (important because this signature is copied in every handler...) as most handlers do not need this param.

tjingrant:
Hi, thanks a lot for your hard work, I've left a few small suggestions here and there; 

One important thing is that tests on GPU are important b/c this is what most people do anyways.

onnx_tf/frontend.py
LucasMahieu: Yes, I agree with you for the signature. It should be betted to keep it without the input node param.
But I did it because I didn't get the idea to add the input shape to the TensorflowNode class.

I will make the fix, thanks

onnx_tf/frontend.py
LucasMahieu: Of course, I will do the helper function as soon as possible.

test/frontend/test_node.py
LucasMahieu: Ok, but with GPU, I think the only correct data format is NCHW, right ? 

Testing all possible data format more dynamically is a good idea.

tjingrant: To quote from this thread https://github.com/tensorflow/tensorflow/issues/2660:

> MaxPool on GPU supports NCHW and NHWC
> MaxPool on CPU only supports NHWC.

I'd expect the same for most other ops that care about input formats.

LucasMahieu: Can you try on a gpu to be really sur of that ?
Because it could impact my backend modifications in order to manage the data format...

LucasMahieu:
Thanks a lot for your feed back, I will take into account all of them, then make a pull request again

test/frontend/test_node.py
tjingrant: I'll let you know my trials later today but that statement was made by TF developers from Google so I'd consider that more authoritative than even my experimental results.

tjingrant:
Hi, there's no need to make another pull request, you can simply continue committing to this branch and the commits will automatically show up here.

LucasMahieu:
Ho, great !

LucasMahieu:
Here is my update, do not hesitate to comment.

test/frontend/test_node.py
```diff
@@ -94,24 +116,44 @@ def do_test_expected(self):
 ("test_or", tf.logical_or, "LogicalOr", [get_rnd([10, 10], dtype=np.bool_), get_rnd([10, 10], dtype=np.bool_)], {}),
 ("test_pow", tf.pow, "Pow", [get_rnd([10, 10]), get_rnd([10, 10])], {}),
 ("test_pad", tf.pad, "Pad", [get_rnd([2, 3]), [[1, 1,], [2, 2]]], {"mode": "constant"}),
+("test_pad", tf.pad, "Pad", [get_rnd([10, 2, 2, 3]), [[0,0], [2, 2], [2, 2], [0,0]]], {"mode": "constant"}),
```
tjingrant: Test name has to be unique. Otherwise only one of these gets run.

tjingrant:
Looking good, let's merge as soon as I run through all the tests on a GPU!

test/frontend/test_node.py
LucasMahieu: Oh, thanks to notice it to me.
Do you want to make to change before to merge ? 
Or do you want I make the modification ?

tjingrant: Could you modify them to have unique names?

LucasMahieu: I can do it ASAP.

LucasMahieu:
Here is my update, do not hesitate to comment.
I have made some modifications due to bugs discovered thanks to tests that was not executed.

tjingrant:
Fantastic, I hope you find these troubles of working on tests worth going through!

Also, given that nothing has really changed (unfortunately) in the spec repo regarding the data format issues we face, do you mind if I edit your comment here https://github.com/onnx/onnx-tensorflow/issues/31#issuecomment-352019711 and add it to somewhere in the README? This is important information that the user needs to know but is currently undocumented.

tjingrant:
Hi, I tried running tests on my GPU but GraphDef reached 2GB at some point and TF refuses to execute the graph. I think I need to add a tear_down function to the unit test in ONNX spec repo that resets the graph at some point...

LucasMahieu:
Great, unit test is fundamental when developing software like this.
I also really think that it is necessary to warn the user in one way or another of the choice of implementation that I made.

For GPU problem I can't help you because I don't have any GPU ...
Do you think we can have access to GPU with CI travis check during pull request ?

tjingrant:
OK, managed to test almost everything on GPU.

PR #43, opened 2018-01-09
Revert "Add flag data format"

tjingrant:
Reverts onnx/onnx-tensorflow#39


tjingrant:
@LucasMahieu it seems to be quite problematic to assume that every tensor certain ops processes are data tensors. For instance, transpose can be used to transpose kernel weights of a convolution and they are in format (M, C, H, W) in ONNX and (H, W, C, M) in Tensorflow regardless of how data_format changes; your implementation will interfere with these transposition as well.

Please let me know what you think. But for now, let us roll back the changes since they are not ready to be merged.

LucasMahieu:
Ho... bad News.
Can you give me an example of the issue you are talking about ?
To be sure I understood the problem.

I will try to make a fix tomorow.
But the real solution is to convince onnx to support data_format attribute.
It is the only good solution to be a true "exchange format".

what do you think about that ?

tjingrant:
The issue is, as I mentioned earlier, not every tensors are organized in the format (N, C, H, W) even within the domain of vision-based DL models. An example is filter, it's organized as (M, C, kH, kW) regardless of whether the image tensor is channel last or not. And in this case, if the user wants to transpose the filter, the transpose operation should not behave differently if the channel_last flag is set.

tjingrant:
Yes, but data_format attribute would be ideal but understandably the community is going to push back on this b/c no other framework likes NHWC format; also having this choice in the Spec while not supporting this format will surely make some other backends look bad. 

I pushed for an alternative solution here: https://github.com/onnx/onnx/issues/406. This will not be enabled by default and can have the extra benefit of having a way for users to express their intentions for the tensors so that each operations can cross-check whether the dimensions are organized as intended.

Vozf:
Are there any updates on this one?

PR #45, opened 2018-02-03
fix build problems

tjingrant:


fumihwh:
@tjingrant 
I think your commit `fix python3 build issue` already fixed the **'build'** problem.

Currently, the problem becomes to **'test failed problem'**.
There is a lots of ops that need to be implement.
Such as hard_sigmoid, hardmax, softsign, top_k, thresholded_relu, unsqueeze, pow, mean.
And attr problem also occurs. We can not use a simple `onnx_tf_attribute_map` to cover all cases. We should use `onnx_tf_per_op_attr_map` instead.

So I think we should let it merge and fix the test failed problem by other prs. What do you think?

tjingrant:
@fumihwh thanks a lot for working hard on this, let's merge and move forward.

Issue #53, opened 2018-02-21
failed in CI testing


fumihwh:
There are two reasons.

- lack of ops
- result not match

Actually, I am working on them and already done the first one.

tjingrant:
I could look at the 2nd one when I get back and if you have not already done so.
A couple of ops have differing definitions in TF; most are variants of the relu ops. I don't think that matters so I may disable the relevant tests soon.

fumihwh:
@tjingrant 
I am working at https://github.com/onnx/onnx-tensorflow/compare/master...fumihwh:fix-build and almost there.
Add and fix a lots of handlers. Let's what will happen https://travis-ci.org/fumihwh/onnx-tensorflow/branches

tjingrant:
Looks great!

tjingrant:
@fumihwh what seems to be the problem?

fumihwh:
@tjingrant 
For backend, FAILED (failures=5, errors=6, skipped=281). [log here](https://api.travis-ci.org/v3/job/344734835/log.txt) Not a big problem.
Frontend is not reached because of that.

tjingrant:
@fumihwh dont worry about fixing every error, I will work on them once I get back next Wed.

I'm fine with merging new changes as long as it does not increase the number of errors/failures.

PR #55, opened 2018-02-24
fix build failed and support multiple outputs

fumihwh:
#53 
Finally, I fixed all build errors and passed the CI test.
But I am not sure if the fix is optimized.
And I added multiple outputs support.

### DIFF
--- a/onnx_tf/backend.py
+++ b/onnx_tf/backend.py
@@ -613,7 +613,7 @@ class TensorflowBackend(Backend):
       return cls.handle_global_average_pool(node, input_dict)
 
     # 0 = cannot pad zero
-    return cls._pool(node, input_dict, partial(tf.nn.avg_pool, pooling_type='AVG'), 0)
+    return cls._pool(node, input_dict, partial(tf.nn.pool, pooling_type='AVG'), 0)
 
   @classmethod
   def handle_batch_normalization(cls, node, input_dict):
--- a/onnx_tf/backend.py
+++ b/onnx_tf/backend.py
@@ -1085,7 +1085,7 @@ class TensorflowBackend(Backend):
     split = (tf.constant(node.attrs["split"]) if
              "split" in node.attrs else input_dict[node.inputs[1]])
     axis = node.attrs["axis"]
-    return [el for el in tf.split(input_dict[node.inputs[0]], split, axis)]
+    return [list(tf.split(input_dict[node.inputs[0]], split, axis))]
 
   @classmethod
   def handle_sub(cls, node, input_dict):
@@ -1112,7 +1112,7 @@ class TensorflowBackend(Backend):
   def handle_top_k(cls, node, input_dict):
     x = input_dict[node.inputs[0]]
     k = node.attrs["k"] if "k" in node.attrs else 1
-    return [el for el in tf.nn.top_k(x, k=k)]
+    return [list(tf.nn.top_k(x, k=k))]
 
   @classmethod
   def handle_unsqueeze(cls, node, input_dict):
--- a/onnx_tf/backend.py
+++ b/onnx_tf/backend.py
@@ -118,7 +118,6 @@ class TensorflowBackend(Backend):
       "low": "minval",
       "axes": "axis",
       "keepdims": "keep_dims",
-      "axis": "dim",
       "to": "dtype",
   }
 
@@ -233,9 +232,6 @@ class TensorflowBackend(Backend):
       "to": lambda cls, x: cls.type_string_to_tf_type[x],
   }
 
-  output_processor = {
-    "TopK": lambda x: [el for el in x[0]],
-  }
 
   # input_shape, kernel_shape, strides are specified for
   # spatial dims only.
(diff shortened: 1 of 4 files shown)


onnx_tf/backend.py
```diff
@@ -1079,7 +1085,7 @@ def handle_split(cls, node, input_dict):
     split = (tf.constant(node.attrs["split"]) if
              "split" in node.attrs else input_dict[node.inputs[1]])
     axis = node.attrs["axis"]
-    return [tf.split(input_dict[node.inputs[0]], split, axis)]
+    return [el for el in tf.split(input_dict[node.inputs[0]], split, axis)]
```
tjingrant: This looks strange.. is it possible to do `list(tf.split(...))` instead?

onnx_tf/backend.py
```diff
@@ -1102,6 +1108,12 @@ def handle_thresholded_relu(cls, node, input_dict):
     epsilon = 1e-5
     return [tf.nn.relu(x) - tf.nn.relu(tf.sign(alpha - x + epsilon) * x)]
 
+  @classmethod
+  def handle_top_k(cls, node, input_dict):
+    x = input_dict[node.inputs[0]]
+    k = node.attrs["k"] if "k" in node.attrs else 1
+    return [el for el in tf.nn.top_k(x, k=k)]
```
tjingrant: Same as above. Does `list` work?

test/backend/test_node.py
```diff
@@ -510,11 +510,11 @@ def test_slice(self):
 
   def test_split(self):
     split = [3, 3, 4]
-    node_def = helper.make_node("Split", ["X"], ["Z"], axis=0, split=split)
+    node_def = helper.make_node("Split", ["X"], ["Z%i" % i for i in range(len(split))], axis=0, split=split)
     x = self._get_rnd([100]).reshape([10, 10])
 
     output = run_node(node_def, [x])
-    for a, b in zip(output["Z"], np.split(x,np.cumsum(split))[:-1]):
+    for a, b in zip([el for el in output], np.split(x,np.cumsum(split))[:-1]):
```
tjingrant: Same. Would `list` work?

onnx_tf/backend.py
```diff
@@ -585,12 +577,22 @@ def _pool(cls, node, input_dict, pool_func, can_pad_zero):
 
     # By default, do not pad
     pad = "VALID"
-
-    if "pads" in node.attrs.keys():
+    pads = node.attrs["pads"] if "pads" in node.attrs else None
+
+    # Pooling1D
```
tjingrant: Why do we have to special case 1-dimensional pooling?
As per TF documentation here (https://www.tensorflow.org/api_docs/python/tf/nn/pool), pooling op does support 1d pooling. 

Quote from the documentation:
> For N=1, the valid values are "NWC" (default) and "NCW".

test/frontend/test_node.py
```diff
@@ -70,7 +70,11 @@ def do_test_expected(self):
     onnx_graph = convert_graph(tf_graph, output_node)
     onnx_model = helper.make_model(onnx_graph)
     backend_rep = prepare(onnx_model)
-    backend_output = backend_rep.run(onnx_feed_dict)[output_name]
+    backend_output = []
+    for ext_output in backend_rep.predict_net.external_output:
+      backend_output.append(backend_rep.run(onnx_feed_dict)[ext_output])
```
tjingrant: Looks like if we have N outputs, the test will run N times.
Any better way to do this?

onnx_tf/backend.py
```diff
@@ -390,7 +382,7 @@ def onnx_graph_to_tensorflow_net(cls, graph_def):
     # defined tensors so we can have access to the placeholders
     # to feed in input tensors when we run the graph.
     original_input_dict = dict(input_dict_items)
-    output_dict = dict()
+    output_dict = input_dict
```
tjingrant: Is this necessary?
Does not look very reassuring...

For `test_operator_chunk`, no node in graph and input is just output. Just for that case.

onnx_tf/backend.py
fumihwh: `1D` means `NWC` or `NCW`

onnx_tf/backend.py
```diff
@@ -1085,7 +1085,7 @@ def handle_split(cls, node, input_dict):
     split = (tf.constant(node.attrs["split"]) if
              "split" in node.attrs else input_dict[node.inputs[1]])
     axis = node.attrs["axis"]
-    return [el for el in tf.split(input_dict[node.inputs[0]], split, axis)]
+    return [list(tf.split(input_dict[node.inputs[0]], split, axis))]
```
tjingrant: I think we shouldn't need the outer brackets. It should simply be:
`list(tf.split(input_dict[node.inputs[0]], split, axis))`

onnx_tf/backend.py
For reference, 
https://github.com/tensorflow/tensorflow/blob/7c324811f10eaf049735ace7b6a228ccc931d8ff/tensorflow/python/layers/pooling.py#L32

fumihwh:
@tjingrant 
`_compatibility_avg_pool` supports 2D (`NCHW` or `NHWC`) only now.
And the implement is very hard coding to fit the 2D.
How should us deal with this?

tjingrant:
@fumihwh I wrote this one to cover an important test case. This shouldn't be used by the user anyways... Let's leave it there and I'll deal with it later.

tjingrant:
Thanks a lot for your help! Glad to see a passing CI badge again!

Issue #47, opened 2018-02-16
Duplicate source codes


fumihwh:
In `backend.py`, `frontend.py`, I find there are a lot duplicate source code which are also in `common.py`. It is not good for maintenance.
I just want to keep them in `common.py` and remove unneeded one in `backend.py`, `frontend.py`.

When this done, I think I could implement some ops(handlers) supported in onnx but not here yet.

tjingrant:
Yes, indeed, I'm trying to move some of the shared codes from `backend` to `common` but didn't want to break anything. It's probably ultimately a good thing to keep things clean.

Would you care to submit a PR for this?

fumihwh:
Yes, I am trying to do this currently.
I will create a PR when it is done.

PR #57, opened 2018-02-26
Remove duplicate source

fumihwh:
#47 
I've removed "seed": "seed2" attr map because I find `tf.random_normal`, `tf.random_uniform` is using `seed` instead of `seed2`.
So I suppose we don't need this attr map anymore.
Also I've moved "to": "dtype" attr map which is only for `Cast` op to per op one.


tjingrant:
Looks reasonable, thanks a lot for your contribution!

PR #58, opened 2018-02-27
use lower case to get dtype

fumihwh:
As Cast test cases added to ONNX, https://github.com/onnx/onnx/blob/master/docs/Operators.md#Cast
Unlike key in there case is uppercase, onnx-tensorflow's is lowercase.
Convert to lowercase to get dtype in backend.


tjingrant:
Can you change the backend test as well? Thanks!

https://github.com/onnx/onnx-tensorflow/blob/f2febba8fc80e8822e02d8790da05b895f3eecd6/test/backend/test_node.py#L141

tjingrant:
Great thanks a lot!

PR #59, opened 2018-02-27
[WIP] add lstm handler

fumihwh:
I've added lstm handler. But without `reverse` direction support. I don't know how to do that. Any ideas?

### DIFF
--- a/onnx_tf/backend.py
+++ b/onnx_tf/backend.py
@@ -48,6 +48,7 @@ from onnx.backend.base import (
 from onnx import onnx_pb2, helper
 import tensorflow as tf
 from tensorflow.python.client import device_lib
+from tensorflow.python.ops import array_ops
 
 # TODO: allow more flexible placement
 def get_device_option(device):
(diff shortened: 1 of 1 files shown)


tjingrant:
I'm as clueless as you are. I thought only forward / bidirectional makes sense.
The best way to resolve this is maybe to raise an issue in the ONNX spec issue page.

tjingrant:
This patch looks good otherwise, maybe just rebase and merge for now?

fumihwh:
@tjingrant 
Sure. But should we add a TODO to it?

fumihwh:
@tjingrant 
```
    elif direction == "reverse":
      def _reverse(input_, seq_dim):
        return array_ops.reverse(input_, axis=[seq_dim])
      time_dim = 0
      inputs_reverse = _reverse(input_dict[node.inputs[0]], time_dim)
      output, state = tf.nn.dynamic_rnn(cell,
                                        inputs_reverse,
                                        time_major=True,
                                        dtype=tf.float32)
```
I read the `bidirectional_dynamic_rnn` [source](https://github.com/tensorflow/tensorflow/blob/r1.6/tensorflow/python/ops/rnn.py#L324-L451) and have found `reverse` part.
Tried to add a if branch into handler. What do you think?

tjingrant:
@fumihwh sounds good.

tjingrant:
@fumihwh thanks a lot!!

PR #61, opened 2018-03-03
 add momentum and spatial support to bn

fumihwh:


onnx_tf/backend.py
```diff
@@ -517,23 +517,24 @@ def handle_batch_normalization(cls, node, input_dict):
...
-                    UserWarning)
-    if "spatial" in node.attrs.keys():
-      warnings.warn("Unsupported spatial attribute by Tensorflow in "
-                    "batch_normalization. This attribute will be ignored.",
-                    UserWarning)
+    spatial = node.attrs.get("spatial", 1) == 1
+    momentum = node.attrs.get("momentum", 0.9)
+    axis = [0] if spatial else [0] + list(range(2, total_num_dim))
+    mean, variance = tf.nn.moments(x, axis)
+    for i in axis:
```
tjingrant: Just wondering why did you not use explicit_broadcast here?

I found it does not work in some cases. So I write by myself here.
But I fix `explicit_broadcast` by another pr finally.
https://github.com/onnx/onnx-tensorflow/pull/62/files#diff-f565871850f10b4c435ec99014386b77L182

So I prefer you to review that one first, if merged, I will add a fix to this pr.

tjingrant:
@fumihwh looks good! 
Thanks again for your contribution!

PR #62, opened 2018-03-04
Support some ops

fumihwh:
- equal
- depth_to_space
- global_lp_pool
- less
- lp_normalization

### DIFF
--- a/onnx_tf/backend.py
+++ b/onnx_tf/backend.py
@@ -180,11 +180,11 @@ class TensorflowBackend(Backend):
 
   @classmethod
   def _explicit_broadcast(cls, tensor, broadcast_dim=1, total_num_dim=4):
-    if not isinstance(broadcast_dim, list):
-      broadcast_dim = [broadcast_dim]
-
+    if broadcast_dim < 0:
+      broadcast_dim += total_num_dim
+    dims = [broadcast_dim + i for i in range(len(tensor.shape))]
     for i in range(total_num_dim):
-      if i not in broadcast_dim:
+      if i not in dims:
         tensor = tf.expand_dims(tensor, i)
 
     return tensor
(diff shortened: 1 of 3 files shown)


onnx_tf/backend.py
```diff
@@ -671,6 +665,17 @@ def handle_conv(cls, node, input_dict):
   def handle_conv_transpose(cls, node, input_dict):
     return cls._conv(node, input_dict, transpose=True)
 
+  @classmethod
+  def handle_depth_to_space(cls, node, input_dict):
+    x = input_dict[node.inputs[0]]
+    x_rank = len(x.get_shape())
+    support_cuda = cls.supports_device("CUDA")
+    storage_format, compute_format = cls.get_data_format(x_rank, support_cuda)
+    x = tf.transpose(x, perm=cls.get_perm_from_formats(storage_format, compute_format))
```
tjingrant: I think we should do the transpose iff storage_format differs from compute_format.

onnx_tf/backend.py
```diff
@@ -180,11 +180,11 @@ def get_padding_as_op(cls, x, pads):
 
   @classmethod
   def _explicit_broadcast(cls, tensor, broadcast_dim=1, total_num_dim=4):
-    if not isinstance(broadcast_dim, list):
-      broadcast_dim = [broadcast_dim]
-
+    if broadcast_dim < 0:
```
tjingrant: I was hoping to support the case where broadcast_dim is a list.. something like broadcast_dim=[1, 2].

I wonder why the change to move away from such support?

Maybe I missed sth. I think `broadcast_dim` is sth https://github.com/onnx/onnx/blob/master/docs/Operators.md#add metioned here.
So I think it is a start dim, and end dim is `broadcast_dim + len(tensor.shape) - 1`.
I also can do list `boardcast_dim`, because it is just `list(range(broadcast_dim, broadcast_dim + len(tensor.shape)))`.

What do you think?

tjingrant: I think your interpretation makes more sense. Let's keep it this way then.

onnx_tf/backend.py
So you mean if branch is desirable?

tjingrant: Yeah.

tjingrant:
Thanks a lot for your contribution!

You should probably add your name to the list of authors in README next time you work on a PR.

PR #63, opened 2018-03-06
support some ops and add default attr value injection

fumihwh:
- ReduceL1
- ReduceSumSquare

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,3 +55,5 @@ Arpith Jacob (IBM Research)
 Tian Jin (IBM Research)
 
 Gheorghe-Teodor Bercea (IBM Research)
+
+Wenhao Hu (LeapMind)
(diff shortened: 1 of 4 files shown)


tjingrant:
@fumihwh why is specialized handlers needed here?
These ops seem to have already been covered here (https://github.com/onnx/onnx-tensorflow/blob/465a02819ebff33ee014ebed17a1b70f66310fcb/onnx_tf/common.py#L117).

fumihwh:
@tjingrant My fault. I will revert there parts.

tjingrant:
Then it looks more like we need the ability to set default attribute rather than there's a need for re-implementing all these operations?

Optimally we should have as many ops automatically handled as possible. What do you think?

fumihwh:
@tjingrant 
BTW, currently I put `DEFAULT_ONNX_ATTR_PER_OP ` in `common.py`. Is it better to put it just in `backend.py`?

tjingrant:
@fumihwh yeah I think moving it into `backend.py` is probably a good idea. Moreoever, why is handle_reduce_sum_square not automatically handled?

fumihwh:
@tjingrant 
I can’t find corresponding op in tensorflow.

tjingrant:
@fumihwh Oh I see.... sorry I missed that.

PR #64, opened 2018-03-08
support some ops, add test cases and remove duplicate source

fumihwh:
- SpaceToDepth
- Tile

### DIFF
--- a/onnx_tf/backend.py
+++ b/onnx_tf/backend.py
@@ -1113,6 +1113,20 @@ class TensorflowBackend(Backend):
     return [tf.reshape(tf.nn.softmax(x - tf.reduce_max(x)), shape)]
 
   @classmethod
+  def handle_space_to_depth(cls, node, input_dict):
+    x = input_dict[node.inputs[0]]
+    x_rank = len(x.get_shape())
+    support_cuda = cls.supports_device("CUDA")
+    storage_format, compute_format = cls.get_data_format(x_rank, support_cuda)
+    if support_cuda:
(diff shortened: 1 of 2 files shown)


tjingrant:
good catch, must be a copy & past error on my side.

onnx_tf/backend.py
```diff
@@ -1112,6 +1112,20 @@ def handle_softmax(cls, node, input_dict):
 
     return [tf.reshape(tf.nn.softmax(x - tf.reduce_max(x)), shape)]
 
+  @classmethod
+  def handle_space_to_depth(cls, node, input_dict):
+    x = input_dict[node.inputs[0]]
+    x_rank = len(x.get_shape())
+    support_cuda = cls.supports_device("CUDA")
+    storage_format, compute_format = cls.get_data_format(x_rank, support_cuda)
+    if support_cuda:
```
tjingrant: I know I did this in other places, but can you do something in the spirit of 
`storage_format != compute_format` 

just in case TF supports NCHW on CPU in the future.

tjingrant: Second thought, we really are just special-casing storage_format=NCHW and compute_format=NHWC. Maybe we should make this clear and I can work on it after this PR.

fumihwh: To first, even `storage_format != compute_format` (or sth like this) would be a little dirty. I will find a new way to deal with this.
Second, yes, and how is ONNX `NHWC`, data format thing going?

tjingrant: Still no response, I posted a new msg today asking for progress.

Issue #70, opened 2018-03-15
remove consts/inputs after use them as attr in frontend


fumihwh:
In case input in TF which is attr in onnx, such as `handle_reshape`:
```
  @classmethod
  def handle_reshape(cls, node, consts):
    assert node.inputs[1] in consts.keys()
    shape = consts[node.inputs[1]]
    return helper.make_node("Reshape",
                            [node.inputs[0]],
                            [node.name],
                            shape=shape)
```
After use it as an attr of node `Reshape`, it should not be a `initializer` or `input` anymore.
We should delete it from graph. (remove them from `inputs_proto` and `consts_proto`)

tjingrant:
That's very nice.

Issue #69, opened 2018-03-15
pass additional args to handlers


fumihwh:
Currently we pass `consts` to handlers in frontend, but in some case, we need more to deal with ops.
For example, at least we need input shape to calculate pad size for pooling and conv because `SAME` and `VALID` is deprecated.
So I have to pass all ops' `_output_shape` to handlers.

Then, there are three questions:
1. do we really need additional args
2. if 1 is yes, what should we pass to handlers
3. how to do that in a smart way

tjingrant:
I think for _output_shape, we can just keep it in the proto attribute instead of a parameter to the handler? I feel like this is easier to handle.

fumihwh:
Actually, I need previous node's `_ouput_shape`....

tjingrant:
Ah, so it's really `input_shape`. In this case, yes, let's put it in the kwargs then.

PR #71, opened 2018-03-15
remove consts/inputs after use them as attr in frontend

fumihwh:
#70

### DIFF
--- a/onnx_tf/frontend.py
+++ b/onnx_tf/frontend.py
@@ -6,6 +6,8 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+from itertools import chain
+
 import tensorflow as tf
 import numpy as np
 from onnx_tf.common import (
@@ -98,8 +100,6 @@ class TensorflowFrontend(object):
     # construction time.
     consts = {}
 
-    used_consts = set()
-
     # Sometimes the constants are used as inputs to ops. This list
     # holds initializers that creates global constant tensors available
     # to be accessed by ops as inputs (as oppose to attributes which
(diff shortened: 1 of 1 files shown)


tjingrant:
I think this is a very good idea. But a safer way to do this is through dependency analysis; i.e. Analyzing in the end whether the input is still needed after we done converting. The reason is, what if TF decides to re-use some of the shared constants? You are assuming that all the consts have a reference count of exactly 1 (or less).

fumihwh:
@tjingrant 
You are right.
We should first think about how to pass additional args to handler. #69 
Is `*args` good enough?

tjingrant:
Hi, kwargs is fine and may be even better than what's done before.
However, I think your current implementation still has a problem: you'll remove constant (or input) tensors if they are used as an input to an operation not as an attribute to an operation; which is obviously quite problematic. I assume this is not the final version you want to merge?

fumihwh:
@tjingrant 
No. Actually I remove consts which is used as attr.
Could you point out which one is `they are used as an input to an operation not as an attribute to an operation`?

fumihwh:
And alternative way is create a `Protos` class to hold all protos and process by class method.

tjingrant:
My bad, I could never remember if the filter condition is the condition to keep or to discard...

Let's examine the logic here, you are assuming:
`A tensor is used by an operation as an attribute -> they are safe to be deleted`

This is a bit unsettling. A very unlikely scenario (but perfectly legal) is when a user used a const tensor both as an attribute and as an input to an operation. I believe your code could cause trouble. Instead, I propose using this logic to determine which const tensor to drop:

A const tensor is not an input to any ONNX operation we generated -> it is safe to be dropped.

Besides, your implementation requires each handlers to record accurately which tensors are used; which is a whole other story..

What do you think?

fumihwh:
`A const tensor is not an input to any ONNX operation we generated -> it is safe to be dropped.` sounds good.

So what we need to do is just loop `ops_proto` at last to find which const is not used as input and dropped it, right?

tjingrant:
Yes, I believe so.

fumihwh:
@tjingrant 
FIxed. Looks good now.

onnx_tf/frontend.py
```diff
@@ -182,15 +190,16 @@ def _bin_op(cls, node, onnx_op):
...
 
   @classmethod
-  def handle_logical_or(cls, node, consts):
+  def handle_logical_or(cls, node, **kwargs):
     return cls._bin_op(node, "Or")
 
   @classmethod
-  def handle_pad(cls, node, consts):
+  def handle_pad(cls, node, **kwargs):
+    consts = kwargs["consts"]
```
tjingrant: What's the purpose of this statement?

tjingrant:
Looks good!

PR #74, opened 2018-03-19
Fix reshape

fumihwh:
https://github.com/onnx/onnx/pull/608
`Reshape` will not use `shape` as attr anymore.
They moved it to `input`.


tjingrant:
This looks good, try merging it yourself now (this will make things smoother for you in the future).

fumihwh:
@tjingrant As onnx says they are using upset things, test should not be skipped for passing ci. 
I will try that first. Just pending to merge.

PR #76, opened 2018-03-19
fix too frequent warnings about kernel_shape

tjingrant:


tjingrant:
Merging b/c fails were all about reshape errors.

PR #77, opened 2018-03-20
preliminary work on adding the optimizer

tjingrant:


onnx_tf/frontend.py
```diff
@@ -183,6 +185,21 @@ def tensorflow_graph_to_onnx_graph(cls, graph_def, output, name="graph"):
                       output_proto,
                       consts_proto)
 
+  @classmethod
+  def tensorflow_graph_to_onnx_model(cls,
+                                     graph_def,
```
fumihwh: Never be called, is it ok?

tjingrant: I imagine that this function is left for users to call.
And it looks like everyone else is doing a command-line interface for model conversion so I'm considering making a command line wrapper to call this function.

But let me export this function nevertheless.

tjingrant:
Looks like the optimizer is not in pip yet.

houseroad:
@tjingrant @fumihwh the optimizer is in the onnx 1.1.0 release now. Please contribute your awesome passes to onnx/optimizer. :-)

PR #78, opened 2018-03-20
generate opset version

tjingrant:


tjingrant:
So this code tries to import backend_v1.py, backend_v2.py ...etc and looks for whether handler exists. If exists, it will record each version where a handler exists. (you can try this out by copying backend.py to backend_v1.py and run this script).

tjingrant:
In addition, I do not see a case where frontend opset version is needed... Do you?

tjingrant:
Merging b/c CI is currently not pertinent to this code snippet.

Issue #75, opened 2018-03-19
support opset


fumihwh:
ONNX is using `opset` to version ops, which is not supported in onnx-tensorflow.
Unless support opset here, we will loss a lot of compatibilities.
One idea is separate handlers to another files - make one file per op.
In file, we have both frontend, backend versioned handlers.
For example,
```
class AddHandler(...):
   
   opsets = [1, 3]
  
   def __init__(self, opset, ...):
      all_opsets = order(opsets + [opset])
      opset = all_opsets[len(all_opsets) - 1 - all_opsets[::-1].index(opset) - 1]

   def get_backend_handler():
      return getattr(cls, 'backend_' + opset)

   def get_frontend_handler():
      return getattr(cls, 'frontend_' + opset)

   def backend_1():
      .....
   
   def backend_3():
      ....

   def frontend_1():
      .....
   
   def frontend_3():
      ....
```

How about this?

tjingrant:
I suspect this will create way too many files and it could be very difficult to manage in the end. I think we should probably do one version per file (and only implement the handlers that have changed since the last version) and by default automatically pick the latest version of the implementation. 

What do you think?

fumihwh:
You mean `backend_2.py`, `backend_3.py` <-- sth like this?

tjingrant:
Yes and we only write ops that have updated definitions in newer versions. So backend_2.py, backend_3.py should be fairly small in size. Also we can pull their shared code to backend_common.py.

fumihwh:
If use this way, it's a bit harder to find appropriate handler.
For example, graph `opset=4` and newest `AddHandler` is `opset=1`.
We should loop `backend_*.py` to check if `AddHanlder` exists.
Should we add a version management file (dict) to manage ops version?
```
AddHandler: [1, 2]
```

tjingrant:
Indeed I think that might be a good idea. 
I think we can possibly generate this file automatically.

fumihwh:
`this file` means indexing file of ops, right?
If yes, sounds good.

tjingrant:
Yes.

fumihwh:
If you have time, otherwise I can try.

tjingrant:
Yes, let's divide the work for this one. So you can assume that you already have a dictionary like 
`
opset = {'add': [1, 2], 'subtract': [2, 3]}
`
where the list of numbers indicate the available versions in our implementation (assuming we are consistent between frontend and backend).

PR #79, opened 2018-03-20
Opset support

fumihwh:
#75


fumihwh:
Maybe should create a `backends` folder and put versioned backend in.

fumihwh:
Backend looks good but frontend does not work.
Maybe we should create an tf version's opset which is matching with onnx opset.

onnx_tf/backend.py
```diff
@@ -337,34 +316,52 @@ def onnx_initializer_to_input_dict_items(cls,
...
-      method_to_call = getattr(cls, handler_name)
+    versions = backend_opset_version[op_name_lowered]
+
+    if opset == 0:
+      version = max(versions)
+    else:
+      versions = sorted(versions + [opset])
+      version = versions[max([i for i, v in enumerate(versions) if v == opset]) - 1]
+
+    backend = importlib.import_module('onnx_tf.backends.backend_v{}'.format(version)).TensorflowBackend
```
tjingrant: I think we should load and cache all versions of the backend as a class variable. Otherwise having a disk access every time we convert an op is not very efficient.

onnx_tf/backend.py
```diff
@@ -337,34 +316,52 @@ def onnx_initializer_to_input_dict_items(cls,
...
+
+    if handler_name in dir(backend):
```
tjingrant: nit: hasattr is preferred as dir is primarily a diagnostic tool.

tjingrant:
This is looking great.
I just have a small question, is frontend_opset_version ever used?

fumihwh:
Never. Just for reference, show which ops we support in frontend.

tjingrant:
Also let's use a squash and merge in the future for keeping a concise commit history.

PR #80, opened 2018-03-20
tiny refactoring

fumihwh:
In case that multiple tf ops match same onnx op.

### DIFF
--- a/onnx_tf/backend.py
+++ b/onnx_tf/backend.py
@@ -352,11 +352,8 @@ class TensorflowBackendBase(Backend):
       version = versions[max([i for i, v in enumerate(versions) if v == opset]) - 1]
 
     backend_ver = 'backend_v{}'.format(version)
-    if backend_ver not in cls.backend_version_cache:
-      backend = importlib.import_module('onnx_tf.backends.' + backend_ver).TensorflowBackend
-      cls.backend_version_cache[backend_ver] = backend
-    else:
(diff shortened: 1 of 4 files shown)


tjingrant:
I feel like this one-to-one condition is something that should be preserved (for error prevention). I think you want to loose this one-to-one check b/c of BiasAdd right? In that case, a special handler for BiasAdd is more appropriate as per this post: https://stackoverflow.com/questions/43131606/whats-the-difference-of-add-methods-in-tensorflow

fumihwh:
Yes, I add `to_list` attr for same reason as you said to make it explicitly.
A special handler for `BiasAdd` makes more sense, but in frontend, do we need to check dtype?
Of course `BiasAdd` is just a example, so the problem is how to define which op should use special handler, which op could mapping directly.

tjingrant:
@fumihwh my take is that in the onnx->tensorflow direction we'll just use add because BiasAdd is baked into the handle_conv handler of the backend. And in tensorflow->onnx direction they are naturally handled different b/c Add has a trivial handler and BiasAdd has a special handler. The problem with the proposed change in common.py is that now we cannot detect unexpected many-to-one mapping in the future in `TF_OP_STR_TO_ONNX_OP`, which should never happen (except for very rare cases like this).

I think BiasAdd is a very rare exception and we should not change the current error prevention mechanism just for this single exception.

fumihwh:
@tjingrant OK～Let's remove that part.

PR #81, opened 2018-03-20
generate documentation

tjingrant:


onnx_tf/doc/support_status.md
```diff
@@ -0,0 +1,233 @@
...
+Backend
+______
+
+| ONNX Op        | Supported Version  |
+| -------------- |:------------------:|
+|unsqueeze|1|
+|arg_min|1|
+|less|1|
+|lp_normalization|1|
+|shape|1, 2, 3, 4, 5|
```
fumihwh: Not relate to this pr, does `1, 2, 3, 4, 5` make user a little confused?

onnx_tf/frontend.py
```diff
@@ -201,6 +202,38 @@ def tensorflow_graph_to_onnx_graph(cls, graph_def, output, opset=0, name="graph"
...
+
+    This function converts a Tensorflow Graph proto to an equivalent
+    representation of ONNX model.
+
+    :param graph_def: Tensorflow Graph Proto object.
+    :param output: A Tensorflow NodeDef object specifying which node
+      to be taken as output of the ONNX graph.
+    :param opset: Opset version of the operator set.
+      Default 0 means using latest version.
+    :param producer_name: The name of the producer.
```
fumihwh: A little question, what does producer do?

tjingrant: Its the name of framework/tool that generated this model.

onnx_tf/doc/support_status.md
tjingrant: true. Maybe I'll fix the way trivial ops are handled.

onnx_tf/frontend.py
fumihwh: I see. User can change this to any name they want.

onnx_tf/opset_version.py
```diff
@@ -1,249 +1,249 @@
...
-    'tanh': [1, 2, 3, 4, 5],
-    'thresholded_relu': [1],
-    'tile': [1],
-    'top_k': [1],
-    'transpose': [1, 2, 3, 4, 5],
-    'unsqueeze': [1],
-    'upsample': [],
-    'xor': [1]}
+    u'a_ten': [],
+    u'abs': [1],
```
fumihwh: I see you always output unicode string.
Is it affected by certain local environment setting?

onnx_tf/frontend.py
tjingrant: Yes, I was hoping to allow users to specify this field if they use our tool as an API and they want to give credit to themselves. I don't think we should allow user to modify it if user generates anything using our command line tool.

onnx_tf/opset_version.py
tjingrant: good catch, i casted the keys of dictionary to strings and now it produces string.

PR #83, opened 2018-03-21
Fix pool ops

fumihwh:
Rewrite `_compatibility_pool ` method.


tjingrant:
Hi, I actually wrote the pooling function just to cover the case of average pooling with `SAME_UPPER` auto pad attribute. Max pooling with SAME_UPPER auto pad attribute does not need such complications because you can use tf.pad and then perform the pooling with `VALID` padding.

fumihwh:
`tf.pad` will pad 0 which will get unexpected result if all numbers in window is negative.

tjingrant:
@fumihwh to be fair, you can pad a very negative number...

tjingrant:
@fumihwh let's keep your solution since it's more logical (and consistent with avg pool) than tf.pad with a very negative number.

fumihwh:
@tjingrant Oh...maybe you misunderstood.
Just auto_pad=SAME_LOWER or None will call `_compatibility_pool`. 
`SAME_UPPER` and `VALID` call `tf.nn.pool`.

tjingrant:
Sorry my typo, in max_pooling with SAME_LOWER, we apply the explicit padding using a very negative number, something like -9999.999 (or FLOAT_MIN to be more accurate) and expect the results to be like real pads in max_pooling op.

tjingrant:
Python 2 keeps having trouble with vgg16 memory allocation. Trying a 3rd time.

tjingrant:
I'm fine with merging since python 3 works and previous logs suggest that vgg16 was the one to blame.

tjingrant:
Great, succeeded on the 4th trial. Let's hope it won't happen again.

PR #87, opened 2018-03-23
update batch norm to opset 6

fumihwh:


tjingrant:
Any changes you made against v1?

fumihwh:
Nop, so I call handler from v1 instead.

tjingrant:
Fantastic.

Issue #84, opened 2018-03-21
code style


fumihwh:
I know in README, we recommend pylint with tensorflow code style.
But in code, indent with 4 spaces also exist......
We should unify it. A simple way is use [`autopep8`](https://github.com/hhatto/autopep8). Although it is different with tensorflow code style.

tjingrant:
Thats a good idea.

fumihwh:
@tjingrant I'll apply autopep8 to all file after merge your refactor pr.

tjingrant:
Sorry I missed that it's different with tf code style...
I'd actually like to maintain that b/c one day we might be merging the frontend to TF (although unlikely b/c we tried once lol).

fumihwh:
I think we can use autopep8. Just need to change space size to 2.
https://www.tensorflow.org/community/style_guide

fumihwh:
From https://www.tensorflow.org/community/style_guide,
I think we can apply autopep8 with space size 2.

Actually, I tried in my local.
pylint score is improved from 6.15/10 to 8.98/10 with tensorflow/tensorflow/tools/ci_build/pylintrc.

fumihwh:
Use yapf with --style='{based_on_style: chromium, column_limit: 79, indent_width: 2}' got 9.14/10.

PR #90, opened 2018-03-26
apply yapf

fumihwh:
#84 
Arguments: `-rip --style="{based_on_style: google, indent_width: 2}"`


onnx_tf/third_party/get_info.py
```diff
@@ -23,45 +23,45 @@
 
 
 def trim(docstring):
```
tjingrant: Please exclude this file from linting. As the name of the folder suggests, modifying this file carries complicated legal implications I do not want to get involved with.

tjingrant:
Please see my only comment. Otherwise looks great!

fumihwh:
@tjingrant Fixed.

PR #91, opened 2018-03-26
fix opset_imports can not be set properly bug

fumihwh:
Need to use OperatorSetIdProto

### DIFF
--- a/onnx_tf/frontend.py
+++ b/onnx_tf/frontend.py
@@ -24,7 +24,7 @@ from onnx_tf.common import (
     op_name_to_lower,
 )
 from onnx_tf.opset_version import frontend_tf_opset_version
-from onnx import defs, helper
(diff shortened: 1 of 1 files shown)


tjingrant:
Hi, this seems to me the proper way of doing it:
https://github.com/onnx/onnx/blob/1a067bac0350b12d0e55ddfc42025fec3474972e/onnx/test/checker_test.py#L130

onnx_tf/frontend.py
```diff
@@ -24,7 +24,7 @@
     op_name_to_lower,
 )
 from onnx_tf.opset_version import frontend_tf_opset_version
-from onnx import defs, helper
+from onnx import defs, helper, OperatorSetIdProto
```
tjingrant: Maybe remove this import?

fumihwh: Sure

PR #93, opened 2018-03-27
support fused_batch_norm and shape in frontend

fumihwh:


tjingrant:
OK, I'll update the tutorial today if everything works out.

Issue #95, opened 2018-03-28
Conv Kernel Shape


tjingrant:
@fumihwh I'm working on the tutorial but it seems like we have some problem recognizing the filter kernel size in conv2d. 

https://github.com/onnx/onnx-tensorflow/blob/552a87387d93b878bdf79ca25896173a74f14aa4/onnx_tf/frontends/frontend_v1.py#L84

Tensorflow filter size is specified as such:
[filter_height, filter_width, in_channels, out_channels]

Thus seems like we should be taking the first d elements instead of the current approach. Maybe we should change it?

fumihwh:
@tjingrant 
https://github.com/onnx/onnx-tensorflow/blob/552a87387d93b878bdf79ca25896173a74f14aa4/onnx_tf/frontend.py#L189-L192
I transposed kernel first.

tjingrant:
Ah, so I explicitly created the variable and the kernel is not present in the variable name. This is actually quite common if one intends to share the same weight between many layers.

In addition, we should have talked about this during code review for an entirely different reason: if we create a scope named "kernel" in TF then every constant in this scope will be transposed by this piece of code...

Basically I don't think this is the right way to capture convolution kernel... What do you think?

tjingrant:
So I managed to get mnist working (with some ad-hoc fix for the kernel transpose). Turns out bias_add needs more hint about broadcast axis and a fix is already submitted.

The reason why I gave up ResNet is that TF ResNet is trained in NHWC mode. I'll have to think about how to import NCHW based models (it's quite complicated because there is only NCHW conv in ONNX). Please comment if you have any thought on this issue.

fumihwh:
Sorry, I can't get it.
Export TF ResNet to onnx failed because of there is only NCHW conv in ONNX?

tjingrant:
I wouldn't say it failed but to get it working more (non-trivial) work is needed. 
Last time I checked, there is no NHWC conv in ONNX.

fumihwh:
From pr's comment,
I think change the consts directly and put transposed kernel into onnx graph (as initializer) is a better way.

tjingrant:
Yeah but here it's not about the kernel that we need to transpose. It's the input feature map that we need to transpose. It's an issue different from the one the PR tries to address.

fumihwh:
Now I got. Transposing input at first and output at last (also with NCHW conv and other ops) should work?

tjingrant:
[...]

PR #96, opened 2018-03-28
[WIP] fix conv kernel problem

fumihwh:
#95 
- get kernel shape in handler

If we need a kernel name list and transpose consts_protos' value before pass to `make_graph` method?
or pass dict of `consts_proto` to handler and change the values? Maybe the latter one.

### DIFF
--- a/onnx_tf/frontend.py
+++ b/onnx_tf/frontend.py
@@ -185,12 +185,6 @@ class TensorflowFrontendBase(object):
       elif node.op == "Const":
         const_dim = len(node.attr["value"].shape)
 
-        # Weight format is MC(HW) in onnx which is (HW)CM
-        if "kernel" in node.name:
-          dims = list(range(np.ndim(node.attr["value"])))
-          node.attr["value"] = np.transpose(
-              node.attr["value"], axes=dims[-2:][::-1] + dims[:len(dims) - 2])
-
         consts[node.name] = node.attr["value"]
         raw_values = ([node.attr["value"].tolist()] if const_dim == 0 else
                       node.attr["value"].flatten().tolist())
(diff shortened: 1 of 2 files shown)


tjingrant:
Yeah I think something like the latter. Basically we transpose it in the conv handler and use the transposed kernel, leaving the original constant unmodified.

fumihwh:
You mean we don't need to put transposed kernel into final graph?
Other backend will not be able to run this graph.

tjingrant:
So the pseudo code would look like:

handle_conv:
    transpose_op = make_onnx_transpose_node( input[1], output_name="transposed_kernel" )
    conv_op = make_onnx_conv_node(input[0], "transposed_kernel")
    return [transpose_op, conv_op]

So I presume it puts the transpose operation into the final graph, but not the transposed value.

fumihwh:
Oh, it never comes to my head.
I think change the consts directly and put transposed kernel into onnx graph (as initializer) is a better way.

tjingrant:
OK but then you have to ensure that the weights are only used as weights, not any other purposes. (We've discussed a similar situation before). Basically you can declare a tensor, use it as the weight of a conv op and then use it somewhere else as a regular data tensor. This is unlikely but perfectly legitimate with respect to the ONNX specification. In this case you cannot change the content of the const tensor because the other operation using the tensor as a regular data tensor will be messed up.

fumihwh:
Another way is create a copy of kernel and transpose it.
Well, it will also make things different.

tjingrant:
I think that might work. What's your concern behind 

> Well, it will also make things different.

?

fumihwh:
training in onnx? maybe.... although we all do not know when it comes....
size should not be a big problem, should it? (double number of weights..)

fumihwh:
@tjingrant 
I tried the way I proposed.
Works on mnist (tf_nchw -> onnx -> caffe2).

PR #97, opened 2018-03-28
Bias Fix

tjingrant:


fumihwh:
Plz remove  BiasAdd in `DEFAULT_TF_ATTR_PER_OP` if you have time.

tjingrant:
@fumihwh done.

fumihwh:
Thanks.

onnx_tf/frontends/frontend_v1.py
```diff
@@ -56,7 +56,8 @@ def handle_fused_batch_norm(cls, node, **kwargs):
 
   @classmethod
   def handle_bias_add(cls, node, **kwargs):
-    return cls._bin_op(node, "Add")
+    data_format = node.attr.get("data_format", "NHWC")
+    return cls._bin_op(node, "Add", axis=data_format.find("C"))
```
fumihwh: should concern 1d and 3d

tjingrant: Yes, but the documentation says default data format is "NHWC". It also says that bias is:

A 1-D Tensor with size matching the last dimension of value.

So I really can't make too much sense out of it..

fumihwh: Sorry I missed that.
It was a shock to me that bias_add only support 2d.....

tjingrant: I think the documentation is wrong and inconsistent in many ways. bias_add certainly supports 3d (c.f., https://gist.github.com/charlesdckim/4833c7a461b8d879ba04034d6619b47c). But the documentation says only "NHWC" and "NCHW" are supported. I think we should adhere strictly to the documentation and submit a change to TF documentation first and then fix it here.

fumihwh: In `gen_nn_ops`, there is hard coding of data_format set default to `NHWC`.
No idea how gist works... Just leave it first.

tjingrant: Yeah. It could be that the logic is:
1. NHWC -> broadcast at -1
2. NCHW -> broadcast at 1

Then works both ways. Just need to update the TF documentation.

fumihwh: So hard to find the implement in TF......

PR #98, opened 2018-03-28
when -1 use None in shape to create placeholder

fumihwh:


tjingrant:
Looks good.

tjingrant:
Although I think to fundamentally address this issue we probably need fix in frontend to use symbolic variable.

fumihwh:
Of course we can fix in frontend to use symbolic variable.
But can not be assured that shape comes with out -1.

tjingrant:
Yes, indeed but my take is that, if we have frontend symbolic variable support, then, if the shape comes out having a dimension size <= 0, this means that we made a mistake somewhere. Thus instead of converting to a reasonable number, we should report an error.

But that being said, this PR is a self-contained fix good for now thus I think we should merge it.

PR #99, opened 2018-03-28
Fix kernel weight + Remove output_shapes

tjingrant:


onnx_tf/frontend.py
```diff
@@ -243,18 +237,16 @@ def tensorflow_graph_to_onnx_graph(cls,
           # Remove tensorflow-specific attrs that are not
           # needed/allowed in ONNX.
           attr = cls.DEFAULT_TF_ATTR_PER_OP.get(node.op, {})
-          attr.update(
-              dict(
-                  filter(lambda pair: pair[0] not in TF_ATTR_TO_REMOVE,
-                         node.attr.items())))
-          node.attr = attr
-          node_output = node.name
+          filtered_attr = dict(
```
tjingrant: Should avoid creating side effects whenever possible. 
Plus, later node translation can make use of previous nodes' attribute as hints.

Another desirable outcome is that, since now we don't overwrite attributes, we don't need to have a dictionary for a specific attribute (like output_shapes).

onnx_tf/frontends/frontend_v1.py
```diff
@@ -81,25 +82,28 @@ def _conv(cls, node, d, **kwargs):
...
+        map(lambda i: node_dict[node.inputs[0]].attr["_output_shapes"][0][i],
+            spatial_indices))
     pads = cls._cal_pads(auto_pad, len(spatial_indices), input_shape,
                          output_shape, strides, kernel_shape)
-    return helper.make_node(
-        "Conv", [node.inputs[0], node.inputs[1]], [node.name],
+    unique_suffix = get_unique_suffix()
+    transpose_node = helper.make_node(
+      "Transpose", [node.inputs[1]], [node.inputs[1] + "_T_" + unique_suffix],
+      perm=[d + 1, d] + range(d))
```
fumihwh: `_T_` for transpose?

fumihwh:
LGTM

onnx_tf/frontends/frontend_v1.py
tjingrant: Yes. 
I'm trying to give more readability to the node in case someone tries to visualize it.

PR #100, opened 2018-03-29
fix frontend API verbosity

tjingrant:
Frontend API for now is too verbose mainly because we have to find the output node oursevles. 

Currently we have:
```
def get_node_by_name(nodes, name):
  for node in nodes:
    if node.name == name:
      return node

with tf.gfile.GFile(graph_pb, "rb") as f:
	graph_def = tf.GraphDef()
	graph_def.ParseFromString(f.read())
	onnx_model = tensorflow_graph_to_onnx_model(graph_def,
                                     get_node_by_name(graph_def.node, "fc2/add"),
                                     opset=6,
                                     producer_name="onnx-tensorflow",
                                     graph_name="graph")
```
Proposing to change to 
```
onnx_model = tensorflow_graph_to_onnx_model(graph_def,
                                     "fc2/add",
                                     opset=6,
                                     producer_name="onnx-tensorflow",
                                     graph_name="graph")
```

### DIFF
--- a/onnx_tf/doc/API.md
+++ b/onnx_tf/doc/API.md
@@ -68,8 +68,8 @@ _params_:
 `graph_def` : Tensorflow Graph Proto object.
 
 
-`output` : A Tensorflow NodeDef object specifying which node
-to be taken as output of the ONNX graph.
+`output` : A string specifying the name of the output
+graph node.
 
 
 `opset` : Opset version of the operator set.
--- a/onnx_tf/frontend.py
+++ b/onnx_tf/frontend.py
@@ -278,8 +278,8 @@ class TensorflowFrontendBase(object):
     representation of ONNX model.
 
     :param graph_def: Tensorflow Graph Proto object.
-    :param output: A Tensorflow NodeDef object specifying which node
-      to be taken as output of the ONNX graph.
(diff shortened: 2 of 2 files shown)


tjingrant:
@fumihwh  This way we will have much cleaner tutorial as well.

fumihwh:
Is there any case that finally outputs is multiple?

tjingrant:
@fumihwh yes I believe that's possible. And it's going to be easy to extend and the extension won't break compatibility (basically we can have a special case where the output is specified as a list). 

This is just for the tutorial..

fumihwh:
Should we make output_node name as an attr? If None, just use graph.node[-1].

tjingrant:
@fumihwh I think that's a good idea in general. But if we do this in the tutorial I'll have to explain why this makes sense (i.e., nodes are topologically sorted) and when can users use it (when you don't have any redundant nodes in TF graph)... We may end up with a more complicated explanation for one less line of code...

Plus, Tensorflow freeze_graph requires the output name to be provided. Mirroring their practice avoids learning overhead. 

But I think it makes sense to add this functionality (just not for the tutorial's purpose..).

tjingrant:
@fumihwh can you approve this PR?

PR #101, opened 2018-03-30
fix range can't concatenate to list bug

fumihwh:


tjingrant:
Good catch. Need to add more frontend tests to catch more Python 3 bugs.

PR #104, opened 2018-03-31
Fix opset imports bug

fumihwh:
domain support does not in this PR. I think we should discuss more about it.

### DIFF
--- a/onnx_tf/frontend.py
+++ b/onnx_tf/frontend.py
@@ -121,7 +121,7 @@ class TensorflowFrontendBase(object):
   def tensorflow_graph_to_onnx_graph(cls,
                                      graph_def,
                                      output,
-                                     opset=0,
+                                     opset={"": 0},
                                      name="graph"):
     """Converts a Tensorflow Graph Proto to an ONNX graph
 
@@ -131,8 +131,7 @@ class TensorflowFrontendBase(object):
     :param graph_def: Tensorflow Graph Proto object.
     :param output: A Tensorflow NodeDef object specifying which node
       to be taken as output of the ONNX graph.
-    :param opset: Opset version of the operator set.
-      Default 0 means using latest version.
+    :param opset: Opset, which should be {str domain: int version number, }.
     :param name: The name of the output ONNX Graph.
 
     :returns: The equivalent ONNX Graph Proto object.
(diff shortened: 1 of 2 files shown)


onnx_tf/frontend.py
```diff
@@ -197,20 +196,20 @@ def tensorflow_graph_to_onnx_graph(cls,
 
         versions = frontend_tf_opset_version[op_name_to_lower(node.op)]
 
+        opset_ver = opset[""]
```
tjingrant: Potentially there are multiple domains right, so should we iterate through the keys?

fumihwh: This PR is just a hot fix.
Multiple domains support need more discussion.
The main reason is I am not sure onnx supports multiple domain.
https://github.com/onnx/onnx/blob/master/docs/IR.md#model-metadata
I will comment on issue.

fumihwh: Maybe I should not add dict interface({'': 6, 'ai.onnx.ml':1}) and keep int in this PR.

tjingrant:
Do you think we should merge this one or wait for clarifications from ONNX standard community?

fumihwh:
@tjingrant 
https://github.com/onnx/onnx/issues/688
ONNX should support multiple domain. I will work on this and make some changes.

fumihwh:
@tjingrant Plz review.

onnx_tf/frontend.py
```diff
@@ -189,40 +188,54 @@ def tensorflow_graph_to_onnx_graph(cls,
...
-              max([i for i, v in enumerate(versions) if v == opset]) - 1]
+              max([i for i, v in enumerate(versions) if v == opset_ver]) - 1]
 
+        camel_domain = "".join(w.title() for w in op_domain.split("."))
         frontend_ver = 'frontend_v{}'.format(version)
         frontend = cls.frontend_version_cache.setdefault(
-            frontend_ver,
-            importlib.import_module('onnx_tf.frontends.' + frontend_ver)
-            .TensorflowFrontend)
+            frontend_ver + "_" + op_domain,
```
tjingrant: Can we add an err msg when the domain is unknown to us?

fumihwh: Sure. I've added NotImplementedError.

PR #108, opened 2018-04-06
add value_info to graph

fumihwh:
Add value_info to graph.
Merits is onnx can use extra information for e.g. optimization. (mainly is shape and type info.)

### DIFF
--- a/onnx_tf/frontend.py
+++ b/onnx_tf/frontend.py
@@ -160,6 +160,10 @@ class TensorflowFrontendBase(object):
     # is supplied by the `consts` map above).
     consts_proto = []
 
+    # This list holds the protobuf objects of type ValueInfoProto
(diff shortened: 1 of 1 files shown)


onnx_tf/frontend.py
```diff
@@ -191,6 +195,13 @@ def tensorflow_graph_to_onnx_graph(cls,
         input_proto = make_tensor_value_info(name, node.attr["dtype"], shape)
         inputs_proto.append(input_proto)
       else:
+        for i in range(len(node.attr["_output_shapes"])):
+          node_name = node.name + ":{}".format(i) if i > 0 else node.name
+          value_info_proto.append(
+              make_tensor_value_info(node_name,
+                                     node.attr.get("T", TensorProto.BOOL),
```
tjingrant: Let's put a TODO somewhere to make sure we remember to put in a proper fix.

PR #111, opened 2018-04-07
refactor bin op

fumihwh:
Refactor some bin ops.
Actually now, just bias add use _bin_op method in frontend.


onnx_tf/frontends/frontend_v1.py
```diff
@@ -16,15 +16,12 @@
 class TensorflowFrontend(TensorflowFrontendBase):
   """ Tensorflow Frontend for ONNX
   """
-
   ONNX_TO_HANDLER = {
```
tjingrant: This is something that needs to be edited manually after new handlers are added right?

If so, can we detect undocumented handlers in gen_opset.py to make sure that we always document handlers this way?

tjingrant:
Re-running CI. Looks good.

onnx_tf/frontends/frontend_v1.py
fumihwh: Yes, you’re right. 
I am thinking about using decorator instead of dict. 
I’ll work on this later.

PR #112, opened 2018-04-08
use decorator to register onnx op

fumihwh:
No more manually maintenance ONNX_TO_HANDLER in frontend.
But detect undocumented handlers in gen_opset.py is not implemented in this PR.


onnx_tf/frontends/frontend_v1.py
```diff
@@ -16,32 +16,14 @@
...
-      "reduce_sum": "sum",
-      "reshape": "reshape",
-      "split": "split_v",
-      "squeeze": "squeeze",
-      "transpose": "transpose",
-      "concat": "concat_v2",
-  }
 
   @classmethod
+  @TensorflowFrontendBase.register_onnx_op("average_pool")
```
tjingrant: Can you import this decorator and thus make it shorter?

onnx_tf/frontend.py
```diff
@@ -409,6 +409,21 @@ def _reduce_op(cls, op, node, **kwargs):
         axes=axes,
         keepdims=node.attr.get("keep_dims", 1))
 
+  @staticmethod
```
tjingrant: Let's leave a comment to document what this function is used for.

PR #116, opened 2018-04-11
fix split of backend

fumihwh:
https://github.com/onnx/onnx/blob/master/docs/Operators.md#Split

### DIFF
--- a/onnx_tf/backends/backend_v1.py
+++ b/onnx_tf/backends/backend_v1.py
@@ -839,9 +839,16 @@ class TensorflowBackend(TensorflowBackendBase):
 
   @classmethod
   def handle_split(cls, node, input_dict):
-    split = (tf.constant(node.attrs["split"])
-             if "split" in node.attrs else input_dict[node.inputs[1]])
+    x_shape = input_dict[node.inputs[0]].get_shape().as_list()
     axis = node.attrs["axis"]
+    if "split" in node.attrs:
+      split = node.attrs["split"]
+    elif len(node.inputs) == 2:
+      split = input_dict[node.inputs[1]]
+    else:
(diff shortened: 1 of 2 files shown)


onnx_tf/backends/backend_v1.py
```diff
@@ -839,9 +839,16 @@ def handle_space_to_depth(cls, node, input_dict):
...
-    split = (tf.constant(node.attrs["split"])
-             if "split" in node.attrs else input_dict[node.inputs[1]])
+    x_shape = input_dict[node.inputs[0]].get_shape().as_list()
     axis = node.attrs["axis"]
+    if "split" in node.attrs:
+      split = node.attrs["split"]
+    elif len(node.inputs) == 2:
+      split = input_dict[node.inputs[1]]
+    else:
+      per_part = x_shape[axis] / len(node.outputs)
```
tjingrant: Do we need "//" here instead of "/"?
c.f. https://stackoverflow.com/questions/183853/in-python-2-what-is-the-difference-between-and-when-used-for-division

tjingrant: Nvm, missed the point here completely.. anyways, do we need any ways to deal with un-even parts?

fumihwh: So I’ve added assert next line. 
Int(per_part) == per_part

tjingrant: Yes, that's why I asked should we accept uneven parts instead of throwing an assertion error as you did with the assertion:

`
Int(per_part) == per_part
`

The spec seems to agree implicitly that number of outputs must divide the axis to split. Tensorflow also explicitly requires this condition to hold. So I think what you did here is correct. Let's merge this one soon.

PR #117, opened 2018-04-11
use fixed axis=1 in biasadd

fumihwh:
axis should always be 1.


tjingrant:
Let's wait on this one. I don't think this is necessarily the case...

fumihwh:
@tjingrant I didn't get your point.
Why there is no necessary to handle bias add?

tjingrant:
Sorry I missed this one.  I mean broadcasting axis is not necessarily always one.
You are assuming that no NHWC formatted tensor will appear in the graph. 

Please refer to our discussion here: https://github.com/onnx/onnx-tensorflow/issues/95.

Intermediate tensors , in case 2 of the scenarios I described, can be in NHWC format.

fumihwh:
I totally forgot that.
The ONNX optimizer mentioned there, if has, how should it works?

PR #118, opened 2018-04-11
enable import onnx_tf.frontend/backend from the outside of repository

tkng:
the wheel package of onnx-tensorflow 1.0 doesn't include `.py` files under `frontends` directory and `backends` directory. (Actually, directories themselves didn't exist.)

This PR fix the issue. In addition, this PR also adds some `import` statement in `onnx_tf/__init__.py`.
The latter change was needed to make it work on my environment. (Python: 3.6.3 with pyenv, OS: ubunu 16.04.)

### DIFF
--- a/onnx_tf/__init__.py
+++ b/onnx_tf/__init__.py
@@ -1,0 +1,2 @@
+from . import frontend
+from . import backend
--- a/setup.py
+++ b/setup.py
@@ -11,5 +11,5 @@ setup(
     author='Arpith Jacob, Tian Jin, Gheorghe-Teodor Bercea',
     author_email='tian.jin1@ibm.com',
     license='Apache License 2.0',
-    packages=['onnx_tf'],
+    packages=['onnx_tf', 'onnx_tf.backends', 'onnx_tf.frontends'],
     zip_safe=False)


tjingrant:
Ok, I can sort of understand why the change to onnx_tf/__init__.py is needed. But why do you need to access the frontends and backends folder? Those folders contain versioned handlers that will be picked up by frontend.py or backend.py automatically; do you want to manage onnx version yourself? I see no advantage of doing that...

tkng:
800b2ca is need to include frontends/backends files into .whl package. I describe the concrete problem in the following lines.

Currently, `pip install onnx_tf` installs onnx_tf 1.0, but it doesn't contain any files under `onnx_tf/frontends` and `onnx_tf/backends`. Since there's no file, `onnx_tf.frontend.tensorflow_graph_to_onnx_model` always fails in `frontend.py` line 244, `importlib.import_module("onnx_tf.frontends." + frontend_ver)`.

As for 5f7f177, I'm not sure this commit is mandatory needed, but in my environment (Python 3.6.3), `onnx_tf.frontend.tensorflow_graph_to_onnx_model` didn't work without the change.

tjingrant:
Hi, after doing some experiments, https://github.com/onnx/onnx-tensorflow/commit/800b2caca69cee7674ad24170af630dcf57ff6a0 seems to be a good idea for our next release. However I cannot make sense of the error you described and you may be mixing development install with pip install. Versioned handlers as you see in `frontends` and `backends` folders are a recent development and we have not uploaded this version to pypi yet. So the problem you described cannot happen because if you install through pip, you would install an early version of onnx-tensorflow that does not require the frontends and backends directory (because at that time we released to pypi, there were **indeed no frontends and backends directory** and the package was correctly uploaded.

That being said, we'll take this pull request because our next release will need your modification. Thanks for your contribution!

tkng:
Sorry for confusing, I thought that git master version and 1.0 has the same interface and internal structure. I guess that the misunderstanding caused the confusion.

In my experiment, 800b2ca is need for `python setup.py bdist_wheel`. Without the commit, any files under `onnx_tf/frontends` and `onnx_tf/backends` would not be included in the wheel package. Hence, as you mentioned, having the commit would be nice from the next version.

Anyway, thanks for merging! I could confirm that now my code works well with git master version.

PR #120, opened 2018-04-13
set default opset to latest

fumihwh:
Forget to set opset to latset.


tjingrant:
Is this an official behavior? Just curious..

Issue #119, opened 2018-04-13
should us use warning instead of raising NotImplementedError in frontend?


fumihwh:
Currently, we raise NotImplementedError in frontend when node.op is not in opset_version.
```
raise NotImplementedError("{} op is not implemented.".format(node.op))
```
Should we use warning instead of error here to make sure pb has been converted even if it can't pass the onnx checker?

PR #122, opened 2018-04-15
make frontend can export unrecognized op with warning message

fumihwh:
#119 
Currently, if tf pb contains op that not implement in our frontends yet (both we haven't impl or onnx haven't defs), the converting will fail.
It's not user friendly. So I let them(ops) pass with warning messages.


tjingrant:
Can we add a flag in our API (something like ignore_unimplemented) and disable that by default? This looks like a very dangerous idea that would help newcomers and hurt experienced developers...

tjingrant:
Why is it narrow? The benefit of doing this in our API is that user can more easily discover this feature. To me, an environment variable is something that is never used in onnx-tensorflow, this makes feature discovery (for user) a bit difficult; they may not see this wonderful feature that could save their day.

fumihwh:
@tjingrant 
First, is not an environment variable, it is an argument with name similar to environment variable.
I am concerning about others dev level features. 
But as you said, how to find them and use them becomes a problem which does not have that priority.
So I will stay with ignore_unimplemented.

onnx_tf/frontend.py
```diff
@@ -310,6 +332,9 @@ def tensorflow_graph_to_onnx_model(cls,
       List or tuple items should be (str domain, int version number).
     :param producer_name: The name of the producer.
     :param graph_name: The name of the output ONNX Graph.
+    :param ignore_unimplemented: Convert to ONNX model with ignoring unimplemented operators.
```
tjingrant: Can we say, 

"Convert to ONNX model and ignore all the operators that are not currently supported by onnx-tensorflow".

Just to be a bit clearer.

fumihwh:
@tjingrant 
I've made some changes and add test for unimplemented op. 
Please review again.

PR #123, opened 2018-04-16
Support opset 6

fumihwh:
Difference between opset 6 and 1 is just abandoning consumed_inputs in almost cases.
I am not sure if this pr is optimal solution.


tjingrant:
Can we try something like this: https://stackoverflow.com/questions/1325673/how-to-add-property-to-a-class-dynamically ?

fumihwh:
There are not all ops in v1 have v6.
And I think apply a special rule to do such thing is not a good idea.

tjingrant:
OK, let's do this:
1. Make a note in this file about the difference.
2. Can you take a look at what's going on with "'real_div': [1, 6, 6]," in the generated opset_version.py?

fumihwh:
@tjingrant 
About real_div, I remove the duplicate handler and fixed now.

Issue #124, opened 2018-04-16
Add support for leaky_rely


achernigin1987:
Currently, it's not possible to convert TF graph to ONNX representation if TF graph contains nodes with _leaky_relu_ activation, because there is no support for tf.math_ops.maximum operator that is used in [TF leaky_relu implementation](https://github.com/tensorflow/tensorflow/blob/92e6c3e4f5c1cabfda1e61547a6a1b268ef95fa5/tensorflow/python/ops/nn_ops.py#L1608).

fumihwh:
@achernigin1987 
You could try my pr if you are hurry.

achernigin1987:
Thanx, it works, but also we need resize_nearest_neigbor operation - I've created the [dedicated issue](https://github.com/onnx/onnx-tensorflow/issues/127) for it

PR #125, opened 2018-04-16
add max op support in frontend and fix backend

fumihwh:
#124


tjingrant:
What seems to be the problem with the current backend?

tjingrant:
I'm guessing problem with 1 tensor input?

fumihwh:
@tjingrant 
Current backend works good now. But I think it is a little tricky.
Use `tf.maximum` seems more straightly and more easy to understand and maintain.
Anyway, I am ok to keep current one.

tjingrant:
The benefit of the current implementation is that you are guaranteed, regardless of the number of input tensors, to have only 2 nodes in your computational graph. This, however, is not the case for your updated implementation and this worries me a bit.

fumihwh:
@tjingrant 
Make sense. Reverted.

PR #129, opened 2018-04-19
remove requires of tensorflow and numpy in setup.py

fumihwh:
@tjingrant 
I know we need tensorflow to do both backend and frontend.
But in practice, not only tensorflow, tensorflow-gpu is also used widely. 
(Actually it's my case) I have to switch environment to do convert.
I think remove them from setup.py and let user install deps by themselves is a better idea.
numpy problem is fixed so remove it also.
What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 ## To install:
 Firstly install [ONNX](https://github.com/onnx/onnx) which cannot be installed by pip unless protoc is available.
 
-Secondly install Tensorflow.
+Secondly install Tensorflow>=1.5.0.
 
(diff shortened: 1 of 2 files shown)


tjingrant:
ok, can we update the README to document we need TF somehow?

I did some googling, it seems like it is difficult to have an either or dependency in setup.py.

fumihwh:
@tjingrant 
One thing should be clarified, always up-to-dated tensorflow is supported?

tjingrant:
I think we can specify 1.5 as minimum (this is the one I used for testing).
Then we bump that version up as we see fit in the future.

PR #132, opened 2018-04-20
fix slice with None in shape

fumihwh:
If None in x shape, np.min will crash.

### DIFF
--- a/onnx_tf/backends/backend_v1.py
+++ b/onnx_tf/backends/backend_v1.py
@@ -789,8 +789,9 @@ class TensorflowBackend(TensorflowBackendBase):
 
     for i in range(slice_len):
       ends[i] = full_sizes[axes[i]] + ends[i] if ends[i] < 0 else ends[i]
-      ends[i] = np.min([full_sizes[axes[i]], ends[i]])
-      starts[i] = np.min([full_sizes[axes[i]], starts[i]])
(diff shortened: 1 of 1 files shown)


onnx_tf/backends/backend_v1.py
```diff
@@ -789,8 +789,9 @@ def handle_slice(cls, node, input_dict):
 
     for i in range(slice_len):
       ends[i] = full_sizes[axes[i]] + ends[i] if ends[i] < 0 else ends[i]
-      ends[i] = np.min([full_sizes[axes[i]], ends[i]])
-      starts[i] = np.min([full_sizes[axes[i]], starts[i]])
+      if full_sizes[axes[i]] is not None:
```
tjingrant: Do we need to emit an error if None is encountered?

fumihwh: Even if None, we can also do slice with out of bounds risk.

fumihwh: @tjingrant Sure.

thatindiandude: Any updates?

tjingrant: I guess this seems to be the right thing to do for now to solve your problem. But let's discuss under your PR to see if there's anything high-level that needs to be done as well.

PR #133, opened 2018-04-20
Add rudimentary LSTM support

stgrue:


onnx_tf/backend.py
```diff
@@ -294,7 +294,7 @@ def onnx_graph_to_tensorflow_net(cls, graph_def, opset):
       for value_info in graph_def.input:
         if value_info.name in initialized:
           continue
-        shape = list(d.dim_value if d.dim_value >= 0 else None
+        shape = list(d.dim_value if d.dim_value > 0 else None
```
tjingrant: I'm curious why this change?

I personally cannot think of any cases where dimension size should be zero. Thus I think somewhere in the graph definition, a mistake is being made and we should not (potentially prematurely) help the user fix the mistake by converting a zero to a None.

On the other hand, there might be legitimate reasons to change (for instance if ONNX makes it clear that dim size = 0 means any number would fit) that I do not know about. Could you elaborate?

stgrue: Yes, sorry, I should have mentioned this. I have asked the ONNX developers about statically unknown dimension sizes in [this thread](https://github.com/onnx/onnx/issues/740), and the way to specify this is to use the `dim_param` attribute. When you do this, the value of `dim_value` automatically becomes 0, so I figured that a 0 represents a dimension of arbitrary size. However, it might be a good idea to ask the developers about this, since this does not appear in the specification (I think).

tjingrant: From your description, using dim_param is **not a necessary condition** for dim_value to be 0. In other words, dim_value=0 does not imply that the user used dim_param. There are still possibilities that users made a mistake. 

So I would suggest you pull an update from the master branch (I think something is being done to check dim_param). And there the corresponding contributor check the root cause (aka dim_param), not a manifestation (aka dim_value) of the problem. Please take a look and see if the update resolves your concern.

stgrue: You are right, the solution from the master branch solves the problem of underspecified dimensions in a much cleaner way. Thank you.

onnx_tf/backends/backend_v1.py
```diff
@@ -581,66 +583,73 @@ def handle_l_r_n(cls, node, input_dict):
...
+
+
+    # Read weights and create TensorFlow LSTMs, differentiating
+    # between the unidirectional and the bidirectional case
+    dummy_session = tf.Session()
+    if direction == "forward" or direction == "reverse":
+      W_weights = input_dict[W].eval(session=dummy_session)
+      R_weights = input_dict[R].eval(session=dummy_session)
+      B_weights = input_dict[B].eval(session=dummy_session)
+      tf_kernel, tf_bias = extract_onnx_lstm_weights(W_weights, R_weights,  B_weights)
```
stgrue: I have noticed another problem with my code, and that is the fact that it implies ranks of the LSTM input tensors (W, B, R) which do not match the ONNX specification. In my code, the tensors have rank 2 (`[4*hidden_size, input_size]`), whereas in the spec, they have rank 3 (`[num_directions, 4*hidden_size, input_size]`).

This problem stems from the fact that I have tested my code using the PyTorch exporter, which seems to generate tensors with incorrect ranks (`num_directions` is ignored). I will check back with the PyTorch team and update my code accordingly. Please do not merge the PR until then.

tjingrant:
You might want to check out this other PR by @fumihwh [here](https://github.com/onnx/onnx-tensorflow/pull/156). Since it's been a while you pushed your last commit...

PR #134, opened 2018-04-22
fix conv transpose in backend

fumihwh:


tjingrant:
Also, I think it's probably a good idea if we document things better so that others can easily understand what's going on. ConvTranspose is one of the messier ops, so this is particularly helpful

fumihwh:
@tjingrant 
Transposed conv docs added.

tjingrant:
I was thinking about inline documentation where we can document as we code. The benefits are really for ourselves mostly because 1 month later you may forget why you did certain things...

tjingrant:
Anyways, feel free to merge once you think that you have enough information to recall what you did and why you did it any time in the future.

PR #140, opened 2018-04-26
Topologically sort the nodes before creating the graph.

akifkamal:
Also add range/gather/greaterequal ops
Topologically sort the nodes before creating the graph.


tjingrant:
Thanks for your contribution, I'm a little bit skeptic about implementing algorithms with this level of computational complexity in Python (especially recursive algorithms). It seems like Python sets stack depth limit of 1000 (c.f. https://stackoverflow.com/questions/3323001/what-is-the-maximum-recursion-depth-in-python-and-how-to-increase-it). So I assume your algorithm crashes with a linear model with >1000 nodes?

akifkamal:
Thanks. Yes Python has recursive limitations and I faced this issue. that is why I used OrderedDict. Nodes in tensorflow are mostly sorted but but not across the same layer. I faced this issue with couple of my conversions where converted model failed because they are not sorted. 

Using OrderedDict will at-least ensure that the order from TF remains intact and we can sort the other nodes which are not sorted.

fumihwh:
Thanks for your contribution again.
There are serval problems in this PR.
- gather op is not version 7 (there is no op version 7 in ONNX yet)
- range/greaterequal ops are not supported by ONNX, so should not use `register_onnx_op`
 An function proto has been introduced for composite op in ONNX, maybe we could use that

And I found https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/grappler/utils/topological_sort.cc , it doesn't work?

akifkamal:
@fumihwh  Thanks.

Sorry, I misunderstood frontend versioning. That is why created the file for v7. I didn't realize it is tied to ONNX version. So will move gather op to the correct version file.

Re: range/greaterequal->I will try function proto. Thanks for letting me know.

For topological sort order, I tried it, but then it means it becomes the requirement for the converter that all models fed to it should be topologically sorted, otherwise generated model will not pass the validation. Currently, there is no validation within the converter for topological sort order, so if I used it with a model that is not sorted, converted model is not usable for me,

fumihwh:
@tjingrant Thanks.
Yes, if could, we should try that first.

akifkamal:
I am unable to verify the tensorflow grappler's python optimizer with training/validation model which topologically sort the nodes.

tjingrant:
@akifkamal could you elaborate why you were unable to verify?

chinhuang007:
Close this PR since we have stopped the frontend support. Thanks for your contribution!

PR #144, opened 2018-04-29
introduce data type caster

fumihwh:
Introduce data type caster in frontend.
Mainly used for casting data type which is different from tf.
For example, [`Tile`](https://github.com/onnx/onnx/blob/master/docs/Operators.md#tile) has type `int64` for `repeats` which could be `int32`, `int64` in [tf](https://www.tensorflow.org/api_docs/python/tf/tile).

### DIFF
--- a/onnx_tf/frontend.py
+++ b/onnx_tf/frontend.py
@@ -306,7 +306,7 @@ class TensorflowFrontendBase(object):
   @classmethod
   def _data_type_caster(cls, protos, data_type_cast_map):
     """Cast to a new data type if node name is in data_type_cast_map.
-    Be used to precess protos to match ONNX type constraints.
+    Be used to process protos to match ONNX type constraints.
 
     :param protos: Target protos.
       TensorProto for inputs and ValueInfoProto for consts.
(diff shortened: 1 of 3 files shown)


tjingrant:
Hi, can you show an example usage? 
Seems like for the current implementation, the variable data_type_cast_map doesn't change.

fumihwh:
@tjingrant 
I will add some codes later.

onnx_tf/frontends/frontend_v1.py
```diff
@@ -225,6 +225,9 @@ def handle_squeeze(cls, node, **kwargs):
   @classmethod
   @register_onnx_op("Tile")
   def handle_tile(cls, node, **kwargs):
+    if kwargs["node_dict"][node.inputs[1]].attr["dtype"] != TensorProto.INT64:
```
tjingrant: You can always set a data_type_cast_map here because you checked if the src data type and target data type are the same in your casting logic.

So you don't really need this "if" here.

onnx_tf/frontend.py
```diff
@@ -294,6 +303,38 @@ def tensorflow_graph_to_onnx_graph(cls,
         initializer=consts_proto,
         value_info=value_info_proto)
 
+  @classmethod
+  def _data_type_caster(cls, protos, data_type_cast_map):
+    """Cast to a new data type if node name is in data_type_cast_map.
+    Be used to precess protos to match ONNX type constraints.
```
tjingrant: precess->process

onnx_tf/frontend.py
```diff
@@ -294,6 +303,38 @@ def tensorflow_graph_to_onnx_graph(cls,
...
+      TensorProto for inputs and ValueInfoProto for consts.
+    :param data_type_cast_map: A {node.name: new_data_type} dict.
+    :return: Processed protos.
+    """
+    if not data_type_cast_map:
+      return protos
+    result = []
+    for proto in protos:
+      new_proto = proto
+      if proto.name in data_type_cast_map \
```
tjingrant: Is it possible to remove this \ and replace it with (, ) around two conditions?

My past experience working on TF shows that TF team doesn't like \.

onnx_tf/frontends/frontend_v1.py
fumihwh: lol... I forgot my own codes....

PR #160, opened 2018-05-10
add if branch to skip no _output_shapes node

fumihwh:


tjingrant:
why is this useful?

fumihwh:
Sometimes node.attr doesn't have "_output_shapes"?
Else, maybe we should add the recommended way to export graph to README.

fumihwh:
@tjingrant 
Yes. At least, we should ask user to export graph by using `add_shape=True`.

PR #162, opened 2018-05-11
add infer shapes to GraphDef

fumihwh:
Preprocess graph_def to add infer_shapes to it anyway.


tjingrant:
Looks like an excellent idea. Would love 1 small change:

Maybe let's not do this by default, since importing the model to the default graph might cause TF to allocate memory for the model, which is bad for huge models. Instead can we detect whether shape info is in the first node and if not, do this?

PR #171, opened 2018-05-14
use native pool as much as possible

tjingrant:


tjingrant:
Still needs work with the pad method recovery. Seems to break tests now.

tjingrant:
There's currently conceptual unclarity in our current implementation. We use None to indicate both padding not specified and padding not compatible, which is confusing. I'm introducing a constant string to indicate the latter situation.

tjingrant:
pr passes latest tests, failing onnx 1.1.0 tests which i believe were wrong.

PR #172, opened 2018-05-15
Support count include pad of pool ops

fumihwh:
`count_include_pad` has been added to onnx.
Although, tf doesn't support that. Use compatibility pool instead.


onnx_tf/backends/backend_v1.py
```diff
@@ -168,6 +188,8 @@ def py_pool(x, kernel_shape, strides_shape, pads, out_shape, pad_shape,
   def _pool(cls, node, input_dict, pool_func, pooling_type):
     x = input_dict[node.inputs[0]]
     x_rank = len(x.get_shape())
+    spatial_size = x_rank - 2
+    strides_shape = node.attrs.get("strides", [1] * spatial_size)
```
tjingrant: I'm curious why is this called `strides_shape` instead of `strides`?

onnx_tf/frontend.py
```diff
@@ -496,7 +496,8 @@ def _pool_op(cls, node, onnx_op, **kwargs):
         onnx_op, [node.inputs[0]], [node.name],
         pads=pads,
         kernel_shape=kernel_shape,
-        strides=strides)
+        strides=strides,
+        **kwargs.get("external_node_kwargs", {}))
```
tjingrant: This is a bit sketchy... The problem with using kwargs is that we won't be able to catch unexpected keyword arguments error... I'd suggest that we do it the conventional way by having another argument with default value, is there a reason against this approach?

PR #184, opened 2018-05-18
[WIP] Refactor frontend

fumihwh:


doc/support_status.md
```diff
@@ -127,55 +127,65 @@ ______
...
-|softmax|1|
-|split_v|1, 2|
-|sqrt|1, 6|
-|squeeze|1|
-|sub|1, 6|
-|sum|1|
-|tanh|1, 6|
-|tile|1|
-|transpose|1|
+|Abs|1, 6|
```
tjingrant: The header for this column is Tensorflow ops, looks lie something has changed? Maybe you are listing supported ONNX operators now?

fumihwh: It's tensorflow ops just before applying name_to_lower.

fumihwh:
@tjingrant Any advise is welcome: `package structure`, new `common`, new `FrontendHandler`, new `OnnxGraph`.......

tjingrant:
From what I can see so far, your pull request is making things much better. Please allow me another 1-2 days to go over it. Thanks for the efforts!

tjingrant:
@fumihwh can you take a look at the CI output here: https://travis-ci.org/onnx/onnx-tensorflow/jobs/381511978 . It seems like some BC breaking changes were introduced in this PR. Specifically we have:

```
TypeError: get_schema() got an unexpected keyword argument 'domain'
```

onnx_tf/handlers/frontend/control_flow_common.py
```diff
@@ -0,0 +1,30 @@
+from onnx_tf.common.broadcast import get_broadcast_axis
+from onnx_tf.handlers.frontend_handler import FrontendHandler
+
+
+class LogicalCommon(FrontendHandler):
+
+  @classmethod
+  def logical_op(cls, node, broadcast=1, **kwargs):
```
tjingrant: It seems like you did you own categorization of ops. In my opinion, a **Taxonomy** (i.e., classification) of ops is very difficult to come up with from scratch and even more difficult to justify and remember. Thus I would recommend that we follow the ONNX's way of categorizing them to remove our intellectual burden of maintaining yet another hierarchical classification of operations. Specifically, `Less Op` is a Logical Op, not a `Comparison Op`  in ONNX (c.f. https://github.com/onnx/onnx/blob/6f4b1b12e56cea736a31411c5f151305f211a729/onnx/defs/logical/defs.cc#L68).

ONNX_VERSION_NUMBER
```diff
@@ -1 +1 @@
-1.1.0
+1.1.2
```
tjingrant: Is this really necessary? We should try to remain as permissive as possible in terms of ONNX version number requirements.

onnx_tf/handlers/frontend/control_flow_common.py
fumihwh: @tjingrant 
I am following tf,
https://www.tensorflow.org/api_guides/python/math_ops
https://www.tensorflow.org/api_guides/python/control_flow_ops
of course, follow onnx is a good idea also.

onnx_tf/common/handler_helper.py
```diff
@@ -0,0 +1,64 @@
...
+import warnings
+
+from onnx import defs
+
+from . import op_name_to_lower
+from onnx_tf.handlers.frontend import *  # noqa
+from onnx_tf.handlers.frontend_handler import FrontendHandler
+
+
+def __get_all_subclasses(clazz, except_regex=None):
```
tjingrant: Are you using `clazz` to avoid name conflict with standard keyword `class`? If so, `class_` is the right way to go according to PEP8 and I quote:
```
single_trailing_underscore_: used by convention to avoid conflicts with Python keyword, e.g.

Tkinter.Toplevel(master, class_='ClassName')
```

ONNX_VERSION_NUMBER
fumihwh: This causes the CI error.
API changed from 1.1.0 to 1.1.2....
I just want to try if it works.

onnx_tf/common/handler_helper.py
```diff
@@ -0,0 +1,64 @@
...
+    all_subclasses = set(
+        filter(lambda c: re.match(except_regex, c.__name__) is None,
+               all_subclasses))
+  return all_subclasses
+
+
+def get_all_frontend_handlers(opset_dict):
+  handlers = {}
+  for handler in __get_all_subclasses(
+      FrontendHandler, except_regex=r'.*Common$'):
```
tjingrant: Can we have some documentations around the definition of FrontendHandler saying that any subclass of FrontendHandler with suffix Common is not treated as an op handler?

TBH, I think your previous way of using register_op decorator was more expressive and friendly in expressing what we want to achieve. Anyways, it's up to you to decide; there's kind of a trade of between the extent of automation and semantic expressiveness.

onnx_tf/common/handler_helper.py
```diff
@@ -0,0 +1,64 @@
...
+      filter(
+        lambda schema: handler.DOMAIN == schema.domain and handler.get_onnx_op() == schema.name,
+        defs.get_all_schemas()))
+    if schemas:
+      since_version = defs.get_schema(
+          handler.get_onnx_op(),
+          domain=handler.DOMAIN,
+          max_inclusive_version=version).since_version
+    else:
+      warnings.warn("Unknown op {} in domain `{}`. "
```
tjingrant: Can you put down a short explanation why should_check should be false?

onnx_tf/handlers/frontend_handler.py
```diff
@@ -0,0 +1,78 @@
...
+
+  @classmethod
+  def get_tf_op(cls):
+    return cls.TF_OP or [cls.__name__]
+
+  @classmethod
+  def get_versions(cls):
+    versions = []
+    for method_name in dir(cls):
+      if method_name.startswith("version_"):
```
tjingrant: Similarly, can we document that functions starting with "version" keyword is treated specially?

onnx_tf/handlers/frontend/control_flow_common.py
tjingrant: I think following TF is probably fine. Let's keep it this way then.

tjingrant: BTW, can we document that here https://github.com/onnx/onnx-tensorflow/blob/master/onnx_tf/doc/IMPLEMENTING_NEW_OP.md ? We need an update there anyways.

PR #186, opened 2018-05-19
simplify get_tf_pad and use native pool when count_include_pad is 1

fumihwh:
- simplify `get_tf_pad` use `_pool_get_shapes`
- use tf pool when `count_include_pad` is 1


onnx_tf/backends/backend_v1.py
```diff
@@ -93,25 +92,43 @@ def _get_output_shape(auto_pad, input_spatial_shape, kernel_spatial_shape,
...
+    return out_shape, new_pads
+
+  # input_shape, kernel_shape, strides are specified for
+  # spatial dims only.
+  @classmethod
+  def get_tf_pad(cls, input_shape, kernel_shape, strides, pads):
+    assert pads is not None
+    num_sp_dim = int(len(kernel_shape))
+
+    if pads == [0] * num_sp_dim * 2 or pads is None:
```
tjingrant: I had `assert pads is not None` right above...

fumihwh: ..just copy paste the old one.....

Issue #185, opened 2018-05-19
export of Reshape from onnx to tensorflow does not handle copied dimensions correctly


anderspapitto:
In ONNX, a 0 in the shape input means "copy this dimension from the corresponding dimension of the input's shape", rather than "give me a dimension which is literally of size 0". However resize is naively translated to a tensorflow operator that has the latter semantics rather than the former.

tjingrant:
I believe this issue has been properly resolved.

PR #187, opened 2018-05-19
Handle dim_size==0 in reshape's shape param

tjingrant:
As per https://github.com/onnx/onnx-tensorflow/issues/185.


tjingrant:
@fumihwh The redundancy of the code is probably better resolved once we refactor the backend like the frontend.

PR #193, opened 2018-05-24
Support ver 7

fumihwh:
I am not sure to use `_limited_broadcast` and `_np_broadcast` is a good idea.
They changed broadcast mechanism to as same as numpy from since version 7.

Also fix the tile version, from 1 to 6. 1 is not implemented yet.


tjingrant:
Huh, so if I understand correctly, since TF conforms with NP style broadcasting, we don't need to do anything special about broadcasting in version 7 right?

tjingrant:
On a separate note, can you comment why we should use cls.make_node instead of the helper.make_node from ONNX (i.e., here https://github.com/onnx/onnx-tensorflow/blob/de00b46c16e5e5ea9507e81eac156a1701034e98/onnx_tf/handlers/frontend_handler.py#L27). I don't have any problem with it, but we may want to document design choice as we go along because we may forget them in the future.

fumihwh:
Yes, we don't need `_explicit_broadcast` to calculate anymore from version7.

PR #195, opened 2018-05-24
Fix conv and identity

fumihwh:


tjingrant:
Mishaps like this make me wonder whether the default values in make_node helper does more harm than good...

My suggestion is that we can have two helpers, one called `make_node` where we build everything from scratch, and no parameter should have default value and the make_node helper needs not take a Tensorflow Node as input.

Another called `make_node_from_tf_node` where we build an onnx node, taking inputs, outputs, names from the tf node given. This way, we separate the good but tedious from the evil but convenient, and we know when to pay extra attention.

PR #199, opened 2018-05-31
Update math_mixin.py

tjingrant:


fumihwh:
@tjingrant 
Please also fix `batch_normalization.py`. Thanks!

PR #202, opened 2018-05-31
Add instance norm

tjingrant:


onnx_tf/backends/backend_v1.py
```diff
@@ -1559,3 +1560,23 @@ def handle_xor(cls, node, input_dict):
   @classmethod
   def handle_greater(cls, node, input_dict):
     return [cls._bin_op(node, input_dict, tf.greater)]
+
+  @classmethod
+  def handle_instance_normalization(cls, node, input_dict):
+    epsilon = node.attrs.get("epsilon", 1e-5)
+    with tf.Session() as sess:
```
fumihwh: why we need sess here?

tjingrant: If that's the case, we don't even need the `.eval()`. But I feel like the problem fundamentally stems from the fact that:
1. You can only use initializer to reliably fill variables in tf layers.
2. There's no way to have an initializer that takes a tf.constant.

Supposedly, to resolve 2, the recommended way is to use tf.get_variable (c.f. https://stackoverflow.com/questions/38805599/why-tf-constant-initializer-does-not-take-a-constant-tensor), but hey we are in tf layers and we can't change the code there because of 1.

Thus, the combination of the above two facts makes situations quite absurd. So I hesitate to make any global changes to our project to accommodate such unreasonableness. It also seems like this is a problem special to using tf layers which currently rarely occur, so I figure local hacks might be more suitable. What do u think?

tjingrant: On the other hand, if beta and gamma are dynamically computed, I'm not sure if it's possible to make use of this instance norm layer.

tjingrant:
@fumihwh I switched to use batch_norm as was done inside Tensorflow. I removed the branches related to NHWC data formats.

fumihwh:
@tjingrant Switch to batch_norm is a good idea. LGTM.

fumihwh (approved):

onnx_tf/backends/backend_v1.py
```diff
@@ -1559,3 +1560,43 @@ def handle_xor(cls, node, input_dict):
...
+
+    inputs = input_dict[node.inputs[0]]
+    inputs_shape = inputs.shape
+    inputs_rank = inputs.shape.ndims
+
+    reduction_axis = 1
+
+    moments_axes = list(range(inputs_rank))
+    del moments_axes[reduction_axis]
+    del moments_axes[0]
```
fumihwh: nit: try to get feature dims? equivalent to list(range(inputs_rank))[2:]?

PR #204, opened 2018-06-04
multi outputs support

fumihwh:


tjingrant:
Can we not change the name? So as to maintain backward compatibility. It also makes sense to use singular noun here because you can say, the output of the graph is a collection of tensors. 

Otherwise everything looks good!

tjingrant:
Seems to have some complaints from frontend tests:
`
Node A is not found in the graph provided
`
c.f. https://travis-ci.org/onnx/onnx-tensorflow/jobs/387581005

fumihwh:
@tjingrant My intuition tells me.... `unicode`...

PR #205, opened 2018-06-04
Refactor backend

fumihwh:


fumihwh:
@tjingrant Take your time :)

tjingrant:
@fumihwh amazing work!

onnx_tf/handlers/backend/control_flow_mixin.py
```diff
@@ -0,0 +1,6 @@
+class LogicalMixin(object):
+  pass
+
+
+class ComparisonMixin(object):
```
tjingrant: Inherit broadcast mixin?

onnx_tf/handlers/backend/reshape.py
```diff
@@ -0,0 +1,44 @@
...
+    indices_gathered = tf.gather(input_shape, copy_indices)
+    indices_scattered = tf.sparse_to_dense(copy_indices,
+                                           tf.cast(tf.shape(shape), tf.int64),
+                                           indices_gathered)
+
+    # Perform the copy wherever requested (wherever dim_size == 0)
+    copied_shape = shape + indices_scattered
+    return [
+        cls.make_tensor_from_onnx_node(
+            node, inputs=[tensor, copied_shape], **kwargs)
```
tjingrant: Why throw in **kwargs? Seem to be creating issues in CI:
TypeError: reshape() got multiple values for keyword argument 'shape'

onnx_tf/handlers/backend/control_flow_mixin.py
fumihwh: wwww, absolutely needed.

onnx_tf/handlers/backend/reshape.py
fumihwh: Because `shape` is in `node.attrs`. And we already pass shape as second input. 
`reshape` not be happy with two shapes....

tjingrant:
@fumihwh I suspect the way you handled attrs might break the fix of reshape:
In particular you have:
```
cls._process_attrs(attrs or copy.deepcopy(node.attrs))
```

When you pass in attrs an empty attribute, this attr will be replaced by node.attrs, c.f. my little experiments below:
```
>>> empty_list=[]
>>> non_empty_list=[1,2,3]
>>> empty_list or non_empty_list
[1, 2, 3]
```
And it's still causing:
```
TypeError: reshape() got multiple values for keyword argument 'shape'
```
in here: https://travis-ci.org/onnx/onnx-tensorflow/jobs/389555565

fumihwh:
@tjingrant also fixed inputs

PR #209, opened 2018-06-09
update ONNX to 1.2.1

fumihwh:


tjingrant:
@fumihwh can I convince you to keep 1.1.2? c.f. https://github.com/onnx/onnx-tensorflow/pull/210. 

For new installations, 1.2.1 will be used automatically by pip, keeping 1.1.2 will simply give users more choice.

fumihwh:
@tjingrant OK to me~
If we want to use following features, we should update.
```
IR Changes
Adds function and attribute reference (PR #802).
Adds dimension denotation (PR #443) and type denotation (PR #879).
```

tjingrant:
Can we use these features in a backward compatible way? I.e., denotations are totally optional, so it should be fine to not use that feature when onnx does not support so. For functions, I suppose this affect only the backend, seems like we only need to recognize them when they appear, in this case optional enablement of this feature based on onnx version number doesn't seem to be too hard either??

I can keep working on the legacy fixes if required; assign them to me and I'll roll out fixes.

fumihwh:
@tjingrant CI of multiple versions is preferred.

PR #210, opened 2018-06-10
some legacy fix

tjingrant:


test/backend/test_onnx_backend.py
```diff
@@ -32,6 +32,21 @@
 if 'TRAVIS' in os.environ:
   backend_test.exclude('test_vgg19')
 
+major, minor, revision = map(int, onnx.version.version.split("."))
+if major == 1 and minor < 2:
```
fumihwh: nit: maybe ~1.1 and opset < 6 can be set to legacy version (add version check return true false)?
appears too many times... 
It's okay to keep it anyway.

onnx_tf/handlers/backend/and.py
```diff
@@ -12,7 +12,7 @@ class Add(LogicalMixin, BackendHandler):
 
   @classmethod
   def version_1(cls, node, **kwargs):
-    return [cls.make_tensor_from_onnx_node(node, **kwargs)]
+    return cls.limited_broadcast(node, **kwargs)
```
fumihwh: THX;)

test/backend/test_node.py
```diff
@@ -121,7 +121,7 @@ def _batch_normalization(self, x, mean, variance, bias, scale,
     return x * inv + (bias - mean * inv if bias is not None else -mean * inv)
 
   def test_batch_normalization(self):
-    if defs.onnx_opset_version() < 6:
+    if legacy_opset_pre_6:
```
fumihwh: .. lack of parentheses

tjingrant: I think I fixed that immediately in the commit named "typo".

PR #212, opened 2018-06-11
enable non-strict mode

tjingrant:


fumihwh:
If in strict mode, the result may be different right?

onnx_tf/backend.py
```diff
@@ -207,22 +214,25 @@ def _onnx_node_to_tensorflow_op(cls,
...
     Convert onnx node to tensorflow op.
 
     Args:
       node: Onnx node object.
       tensor_dict: Tensor dict of graph.
       opset: Opset version of the operator set. Default 0 means using latest version.
-
+      strict: whether to enforce semantic equivalence between the original model
+        and the converted tensorflow model, defaults to True (yes, enforce semantic equivalence).
+        Changing to False is strongly discouraged.
```
fumihwh: Could we list up target op that strict would be applied to?

tjingrant: Good idea, will do soon.

tjingrant:
> If in strict mode, the result may be different right?

Strict=True -> Results are the same.
Strict=False -> Results may be different.

To be honest, strict mode is very very very bad from a technical point of view. Since small deviations in the beginning of a neural network can be augmented layer by layer. But the fact that TF doesn't support arbitrary padding is a headache of its own kind. I hope that we are choosing the right "lesser of two evils".

onnx_tf/backend.py
fumihwh: @tjingrant And detail doc for these ops is more expected......

PR #227, opened 2018-07-19
GRU functions

tengyifei:
1. Support `linear_before_reset` by implementing a different GRU with that option
2. Fix errors using bidirectional GRU


tjingrant:
@fumihwh will probably give better review.
I have two general question/comment:
1. Does GRU from Keras https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/layers/recurrent.py with the `reset_after` attr do what you want?

with mapping between ONNX attribute `linear_before_reset` and keras attribute mapping as:
```
g(Xt*(Wh^T) + (rt . Ht-1)*(Rh^T) + Rbh + Wbh) # default, when linear_before_reset = 0 -> reset before matrix multiplication
g(Xt*(Wh^T) + (rt . (Ht-1*(Rh^T) + Rbh)) + Wbh) # when linear_before_reset != 0 -> reset after matrix multiplication
```
2. It's fine with me to have this implementation here in onnx-tensorflow. But do you feel like maybe it's better to move this under tensorflow?

onnx_tf/handlers/backend/lstm.py
```diff
@@ -195,9 +195,7 @@ def _common(cls, node, **kwargs):
       h_fw = tf.expand_dims(state_fw[1], 0)
       h_bw = tf.expand_dims(state_bw[1], 0)
       h = tf.concat((h_fw, h_bw), axis=0)
-      output_fw = tf.expand_dims(output_fw, 1)
-      output_bw = tf.expand_dims(output_bw, 1)
-      output = tf.concat((output_fw, output_bw), axis=1)
+      output = tf.concat((output_fw, output_bw), axis=-1)
```
fumihwh: Why don't need expand_dims anymore?

fumihwh:
@tengyifei 
Thanks for contribution. 
I agree with @tjingrant . I prefer moving this implementation to tensorflow.

thatindiandude:
Having used these changes in a system which takes in this GRU's output as input to another GRU layer, the output is off from the equivalent PyTorch model by `1e-3`. Do you know why this might be the case? For sanity, I checked to see if a one-layer GRU in PyTorch would translate to this GRU w/ linear before reset and the average difference in the elements was `1e-9`. I'm not sure why this implementation has brought so much trouble to subsequent layers.

onnx_tf/handlers/backend/gru.py
```diff
@@ -173,9 +305,7 @@ def _common(cls, node, **kwargs):
       h_fw = tf.expand_dims(state_fw, 0)
       h_bw = tf.expand_dims(state_bw, 0)
       h = tf.concat((h_fw, h_bw), axis=0)
-      output_fw = tf.expand_dims(output_fw, 1)
-      output_bw = tf.expand_dims(output_bw, 1)
-      output = tf.concat((output_fw, output_bw), axis=1)
+      output = tf.concat((output_fw, output_bw), axis=-1)
```
thatindiandude: This return format has kept subsequent GRU cells from taking on a `GRUCellWithLinearBeforeReset`s output. 

`ValueError: Dimension must be 3 but is 4 for 'transpose_41' (op: 'Transpose') with input shapes: [?,8,2048], [4].`

The old formatting worked instead.

sinkpoint:
I'm running into this issue trying to translate PyTorch -> Tensorflow. Since they have different implementations, this is a roadblock. 
It's unlikely at this point that Tensorflow will include this into its codebase, I vote for merging it here.

tjingrant:
@sinkpoint I'm fine either way, but there is a legitimate concern around the maintenance of such piece of code. Is @tengyifei still around for taking this patch to its conclusion?
 
@fumihwh what do you say?

tengyifei:
@tjingrant Sorry I was distracted by other things! I would love to land this if possible. Personally I don't expect the TensorFlow team to merge the GRU extension into their library. Is there another place for such "adaptors"?

iiSeymour:
This is also an issue for me so it would be great to see it merged here.

chinhuang007:
Close due to inactivity. Please reopen or create a new PR if needed.

PR #228, opened 2018-07-22
support max_pool_with_argmax v8

fumihwh:


onnx_tf/handlers/backend/pool_mixin.py
```diff
@@ -60,13 +61,36 @@ def pool(cls, node, input_dict, pool_func, pooling_type, strict=True):
...
+    if pooling_type in ("AVG", "MAX"):
+      if strict and count_include_pad == 0:
+        if pad is PAD_TF_INCOMPATIBLE:
+          return cls._compatibility_pool(node, input_dict, pooling_type)
+      else:
+        if pads != [0] * spatial_size * 2:
+          x = PadMixin.get_padding_as_op(x, pads)
+        pad = "VALID"
+    elif pooling_type == "MAX_WITH_ARGMAX":
+      if pad in (None, PAD_TF_INCOMPATIBLE):
```
tjingrant: Pad cannot be `None` here.
I'm inclined never to allow pad to take `None` value because the semantics of None is unclear in this context (i.e., does it mean no padding or unspecified padding or incompatible padding?). Therefore I think it's probably better to just write 
`assert(pad is not None)` before checking `pooling_type` just to make it explicit.

fumihwh:
@tjingrant I've added two cases to ONNX.

PR #236, opened 2018-08-16
add pack unpack to frontend

fumihwh:


tjingrant:
Just curious, so this is the first op in the frontend that does not have a corresponding onnx op?

fumihwh:
@tjingrant I didn't find corresponding op of Pack / Unpack in ONNX....

PR #243, opened 2018-09-05
bias_add boradcasting fix

tjingrant:


tjingrant:
bias add cannot correctly broadcast when data_format is NCHW in the latest opset.

onnx_tf/handlers/frontend/bias_add.py
```diff
@@ -24,4 +26,25 @@ def version_6(cls, node, **kwargs):
 
   @classmethod
   def version_7(cls, node, **kwargs):
-    return cls.arithmetic_op(node, **kwargs)
+    data_format = node.attr.get("data_format", "NHWC")
+    channel_first = data_format[1] == "C"
+    axis = 1 if channel_first else -1
+
+    unsqueeze_suffix = get_unique_suffix()
+    if (axis == 1):
```
fumihwh: Just some style check.
redundant parentheses.

onnx_tf/handlers/frontend/bias_add.py
```diff
@@ -24,4 +26,25 @@ def version_6(cls, node, **kwargs):
...
+      # In this case, we manually unsqueeze the bias term
+      # to facilitate broadcasting.
+      num_sp_dim = len(data_format) - 2
+      unsqueeze_axes = [i + 1 for i in range(num_sp_dim)]
+      reshape_node = cls.make_node_from_tf_node(
+          node, [node.inputs[1]], [node.inputs[1] + unsqueeze_suffix],
```
fumihwh: without underscore as an connector?

onnx_tf/handlers/frontend/bias_add.py
```diff
@@ -1,7 +1,9 @@
+from onnx_tf.common import get_unique_suffix
 from onnx_tf.handlers.frontend_handler import FrontendHandler
 from onnx_tf.handlers.handler import onnx_op
 from onnx_tf.handlers.handler import tf_op
 from .math_mixin import ArithmeticMixin
+import copy
```
fumihwh: nit: put build-in library at first and an empty line then third party library then self modules.

onnx_tf/handlers/frontend/bias_add.py
tjingrant: I think we never used underscore as a connector. It was only used to emphasize the T (transposed) here (https://github.com/onnx/onnx-tensorflow/blob/317d70b763ba91a3eb2e017c2783beb9ff4b3b32/onnx_tf/handlers/frontend/conv_mixin.py). Regardless I think you have a point and this should make the output name more readable.

fumihwh: Agree with point of readable.

tjingrant:
Hi @fumihwh the nit has been resolved right?

PR #252, opened 2018-09-14
Frontend model test

tjingrant:
Added end-to-end model level testing from
1. TF checkpoint/graph freezing.
2. Get reference tf output results (X, Y).
3. Export to onnx via onnx-tf.
4. Load the onnx model using onnx-tf backend.
5. Run reference input X, get Y'.
6. Check equivalence of (Y, Y').

Should be comprehensive/sufficient enough to ensure success of exporting.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ setuptools.setup(
     version=version,
     description=
     'Tensorflow backend and frontend for ONNX (Open Neural Network Exchange).',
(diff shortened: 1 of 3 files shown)


tjingrant:
@fumihwh please let me know if there are other places where I abuse parentheses. 
Otherwise merging, CI looks good.

PR #265, opened 2018-10-01
Add Split

fumihwh:


onnx_tf/handlers/frontend/split.py
```diff
@@ -5,33 +5,50 @@
...
 class Split(FrontendHandler):
 
   @classmethod
   def args_check(cls, node, **kwargs):
-    if cls.SINCE_VERSION == 2:
-      if node.inputs[1] not in kwargs["consts"]:
-        exception.CONST_NOT_FOUND_EXCEPT(node.inputs[1], node.op_type)
-    if node.inputs[2] not in kwargs["consts"]:
-      exception.CONST_NOT_FOUND_EXCEPT(node.inputs[2], node.op_type)
+    if node.op_type == "SplitV":
```
tjingrant: Do we really need to do arg check?

My understanding is that, for a specific Split, the schema should be fixed, updates to the op schema will necessitate a new name given to the op (i.e., SplitV2). So instead of checking the validity of the graph ourselves, can we let tensorflow check it all together before we pass it in? Feels like this is a big burden to us that we can possibly pass on to tensorflow.

PR #267, opened 2018-10-01
Updated readme to contain warning for testing

jsigee87:
On my (relatively standard) desktop, the test suite takes 938 seconds, and downloads several very large neural networks. There should be a warning here for this for obvious reasons including hard disk space and network connectivity. Feel free to reject or modify the PR to make the warning more explicit. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ Because users often have their own preferences for which variant of Tensorflow t
 
 ## To test:
 For backend, run `python -m unittest discover test`.
+Warning: This will download several large neural networks and may take several minutes.
 
 ## Example:
 In this example, we will define and run a Relu node and print the result.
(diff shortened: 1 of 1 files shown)


tjingrant:
@jsigee87 good point. I would rephrase it a little bit so as not to deter users from testing the installation on their target (production) environment as one absolutely should. Something like:

```
Testing requires significant resources ..., but nonetheless, we highly recommend that users run through the complete test suite before deploying onnx-tf.
```

And just FYI, it's not "several minutes". The complete test suite takes somewhere from 30 min to 60 min to run through on state of the art machines. The reason is, if you run our tests on a machine with only CPU's, only half of the tests gets run (only those that run on a CPU device). So please update your warning to something more accurate, maybe something like "15 to 45 min depending on hardware configurations".

Tests should be taken seriously and we regularly test them to provide reliability, users should do the same.

PR #274, opened 2018-10-05
add model converter API

fumihwh:


tjingrant:
Maybe we should update the documentation to let users know we have this feature?

chinhuang007:
Shall we have some associated test code for new features?

tjingrant:
Shall we also include another mini-example?
We can drop the example about relu.
```
model, img = onnx.load("model.onnx"), np.fromfile("img")
model_output = onnx_tf.backend.prepare(model).run(image)
```

PR #275, opened 2018-10-05
Fix slice negative starts

chinhuang007:
Add support for slice negative starts, which is documented in ONNX
https://github.com/onnx/onnx/blob/master/docs/Operators.md#Slice
Also add unit test for slice.


tjingrant:
@ibm-jenkins test this please.

tjingrant:
Thank!

PR #282, opened 2018-10-12
Use proper op names in frontend_tf_opset_version

chinhuang007:
Change the operator names in frontend_tf_opset_version so the
list reflects the tf operator names in the graph as well as
the tf_op declared in the frontend handlers, rather than
maps to the tf python API.

This should affect only the clarity of documentation, no impact
to running code.


tjingrant:
@chinhuang007 do we need to re-run the gen_doc?

chinhuang007:
@tjingrant Makes sense to regenerate the doc.

tjingrant:
It is strange that changes from another patch is showing up here in this current diff set; maybe you or someone else did a git pull from another patch? Anyhow, can you try rebasing with master? (c.f. https://stackoverflow.com/questions/7929369/how-to-rebase-local-branch-with-remote-master)

I hope to eliminate the chance that unintended changes may slip through.

chinhuang007:
Thanks for the review. The unintended changes should be removed now.

chinhuang007:
@tjingrant You are right. Just included the generator code and regened the list and doc.

tjingrant:
i may want to add a check in our CI to make sure that gen_doc is up to date. I think i forgot to do that again with strided slice.

PR #285, opened 2018-10-13
support strided slice 

tjingrant:


tjingrant:
@fumihwh updated as per our discussion.

tabsun:
This branch cannot support StridedSlice when begin_mask or end_mask is not 0, right?
If I want to add begin_mask and end_mask, do you have some opinions?

tjingrant:
@tabsun yes, you are right. Looking at tf documentation for begin_mast and end_mast:

> If the ith bit of begin_mask is set, begin[i] is ignored and the fullest possible range in that dimension is used instead. end_mask works analogously, except with the end range.

You can probably just manipulate the starts and ends tensor passed to DynamicSlice ops based on the bit mask value at each position. Several ops I think might be useful: Shape, Gather, Scatter and you can find their documentations in the onnx repository here (https://github.com/onnx/onnx/blob/master/docs/Operators.md).

You can also post your work in progress so that we can discuss as you move forward.

tabsun:
@tjingrant Thanks for your work and suggestions.

PR #286, opened 2018-10-14
refactor handler_helper

fumihwh:


tjingrant:
Can we document the usage of onnx_op and tf_op and what these information will be used? I find it quite important for future developers.

PR #287, opened 2018-10-14
Update shape.py

tjingrant:


tjingrant:
c.f. https://www.tensorflow.org/api_docs/python/tf/shape

fumihwh:
https://github.com/onnx/onnx/blob/master/docs/Operators.md#outputs-75 ?

tjingrant:
Updated version of this patch will be included in the cli test patch.

PR #290, opened 2018-10-15
add transpose nodes to conv if NHWC

fumihwh:


tjingrant:
Just one nit, maybe pb_wrapper instead of pb_class? It's up to you.

PR #292, opened 2018-10-18
Add operator checker for frozen graph file

chinhuang007:
Add a checker to check whether the operators in the
frozen graph file are supported in the current
frontend handlers. The input file can be either binary
(.pb) or text (.pbtxt). All unsupported operators
will be shown in the console output.

I think this is a handy utility for anyone to check against
a frozen graph file. We might want to integrate it into
CLI. So this is still work-in-progress.


chinhuang007:
Added args_check for supported operators and integration with CLI. Please review.

chinhuang007:
Just cleaned up and rebased. @tjingrant  @fumihwh could you take a look to see if it makes sense to add this checker to cli. Thanks.

onnx_tf/opr_checker.py
```diff
@@ -0,0 +1,133 @@
...
+  for node in graph_def.node:
+    node_dict.add(node.op)
+
+  node_dict.discard('Placeholder')
+  node_dict.discard('Const')
+
+  logger.info('There are %s unique operators in the model file.',
+              str(len(node_dict)))
+  for k in node_dict:
+    if k not in opset_version.frontend_tf_opset_version:
```
fumihwh: I don't think should use `opset_version.frontend_tf_opset_version`. It's a generated doc not a code based solution.

onnx_tf/opr_checker.py
```diff
@@ -0,0 +1,133 @@
...
+      unsupported.add(k)
+    else:
+      supported.add(k)
+
+  if unsupported:
+    logger.info(
+        'There are %s operators currently not supported by the ONNX '\
```
fumihwh: `not supported by the ONNX` is not precise.

onnx_tf/opr_checker.py
```diff
@@ -0,0 +1,133 @@
...
+  parser.add_argument(
+      "--infile",
+      "-i",
+      help="Input file path, can be pb, pbtxt, or ckpt file.",
+      required=True)
+
+  return parser.parse_args(args)
+
+
+def check_opr_support(graph_def):
```
fumihwh: Please add doc for method.

onnx_tf/opr_checker.py
```diff
@@ -0,0 +1,133 @@
...
+  ext = os.path.splitext(graphfile)[1]
+  graph_def = tf.GraphDef()
+  if ext == ".pb":
+    with tf.gfile.GFile(graphfile, "rb") as f:
+      graph_def.ParseFromString(f.read())
+  elif ext == ".pbtxt":
+    with tf.gfile.GFile(graphfile, "rb") as f:
+      text_format.Merge(f.read(), graph_def)
+  elif ext == ".ckpt":
+    latest_ckpt = tf.train.latest_checkpoint(os.path.dirname(graphfile))
```
fumihwh: Maybe could reuse this part. Such as `load_graph_from_ckpt`

onnx_tf/opr_checker.py
chinhuang007: I believe opset_version.frontend_tf_opset_version is generated using onnx_tf/common/handler_helper.py get_frontend_coverage(), therefore it is from code, not just documentation.

onnx_tf/opr_checker.py
chinhuang007: The sentence is precise "There are %s operators currently not supported by the ONNX exporter for your model." The supported tf operators are extracted from the frontend handlers' @tf_op("xxx"). If an operator is not declared in any handlers, it is not currently supported by the exporter/frontend. If the comment is about the wording "ONNX exporter" I can change it to "ONNX Tensorflow frontend handler".

onnx_tf/opr_checker.py
chinhuang007: Sure. Will add doc.

onnx_tf/opr_checker.py
fumihwh: ONNX-Tensorflow should be better if you want to use full name.
Seems strange without hyphen...

chinhuang007: Just added hyphen

onnx_tf/opr_checker.py
```diff
@@ -0,0 +1,157 @@
...
+
+  logger.info('We checked %d supported nodes for required arguments.',
+              total_nodes)
+  logger.info('  # of nodes passed the args check: %d',
+              total_nodes - failed_nodes)
+  logger.info('  # of nodes failed the args check: %d', failed_nodes)
+  return failed_nodes == 0
+
+
+def load_graph_from_ckpt(ckpt_file):
```
fumihwh: Good.
I will refactor later. `load_graph_from_ckpt` is also needed in `converter.convert`.

onnx_tf/opr_checker.py
```diff
@@ -0,0 +1,157 @@
+import argparse
+import warnings
+import logging
+import os
+
+import tensorflow as tf
+from google.protobuf import text_format
+
+from onnx import defs
```
fumihwh: nit: Group and order by `bulidin`, `third-party`, `onnx-tf`

onnx_tf/opr_checker.py
tjingrant: hi @chinhuang007 i think using opset_version is clearly not as reliable as using get_frontend/backend_coverage() directly. Maybe we can settle with the middle ground in which we invoke these methods directly instead of relying on opset_version.py, thus making this more towards a code based solution rather than a doc based one.

We can also eliminate the possibility that someone forgot to run gen_opset.py, causing opset_version to be out of sync.

chinhuang007: In my opinion, opset_version should be always up-to-date. We need integrity between doc and code. It is not acceptable to create or change handlers not re-running the regen code. Having said that, I changed the checker to use get_frontend_coverage(). Please review.

tjingrant: > We need integrity between doc and code. It is not acceptable to create or change handlers not re-running the regen code. 

@chinhuang007 i totally agree. Both of you make valid points, i proposed such change to strike a balance between concerns from both of you. Thanks for moving this patch forward.

PR #295, opened 2018-10-18
Add Size frontend handler

chinhuang007:
Add Size frontend handler support. Also add to the
frontend test to ensure the output from sess.run is
iterable, to be consistent with backend_output, before
using it in zip.


onnx_tf/handlers/frontend/size.py
```diff
@@ -0,0 +1,12 @@
...
+from onnx_tf.handlers.handler import onnx_op
+from onnx_tf.handlers.handler import tf_op
+
+
+@onnx_op("Size")
+@tf_op("Size")
+class Size(FrontendHandler):
+
+  @classmethod
+  def version_9(cls, node, **kwargs):
```
fumihwh: Size op is from 1 not 9. 
https://github.com/onnx/onnx/blob/master/docs/Operators.md#version-88

onnx_tf/handlers/frontend/size.py
```diff
@@ -0,0 +1,12 @@
...
+    return cls.make_node_from_tf_node(node, node.inputs[0], **kwargs)
```
fumihwh: Default dtype of tf.size output is int32. We should cast to int64 to match onnx.
https://github.com/onnx/onnx/blob/master/docs/Operators.md#outputs-79

chinhuang007: Thanks for the comments. Changed the version to 1. But I don't know what needs to be done for the output dtype in the frontend handler. The backend handler was written before and already supports the output cast to int64 when calling tf.size. For the frontend handler, can't we simply ignore the dtype in tf.size (as we did for tf.shape) and allow onnx.size to return int64?

fumihwh: We can't allow that. It will break the constraint.
For `shape`, it should be fixed.
Check the attr `out_type`, do nothing if it's `int64`, else, add `Cast` to cast it to `int32`.

chinhuang007: Got it. Added Cast to size frontend handler. Please review. If the code is okay, I can look into shape next.

test/frontend/test_node.py
```diff
@@ -86,6 +87,12 @@ def do_test_expected(self):
       with tf.Session() as sess:
         tf_output = sess.run(test_op, tf_feed_dict)
 
+      # make sure backend_output and tf_output are Iterable
+      if backend_output.ndim == 0:
+        backend_output = backend_output.reshape(1)
+      if isinstance(tf_output, Iterable) == False:
+        tf_output = [tf_output]
+
```
fumihwh: Why do we need this block?

onnx_tf/handlers/frontend/size.py
```diff
@@ -0,0 +1,39 @@
...
+    size_node = cls.make_node_from_tf_node(
+        node, [node.inputs[0]],
+        outputs=[size_output_name],
+        name=node.name + size_suffix)
+
+    if not need_cast:
+      return [size_node]
+
+    dst_t = mapping.NP_TYPE_TO_TENSOR_TYPE[np.dtype(
+        tf.as_dtype(tf.int32).as_numpy_dtype)]
```
fumihwh: Will not work if user want to export an opset v1 model.

test/frontend/test_node.py
```diff
@@ -95,6 +102,7 @@ def do_test_expected(self):
 
   return do_test_expected
 
+
```
fumihwh: Is this added by yapf automatically?

test/frontend/test_node.py
fumihwh: What is the reason that error occurs after this patch? I mean before this, every think works well.

chinhuang007: It is because for size(), backend_output.shape could be (), and tf_output type could be <class 'numpy.int32'>. Neither is iterable.

onnx_tf/handlers/frontend/size.py
chinhuang007: The updated code is supposed to work with opset v1. But I found a logic flaw in determining the frontend handler.SINCE_VERSION. In this case, size is in ONNX opset v1 and never changed, while cast is in ONNX opset v1 and changed in v6. Now I need to use version_6 in my handler when opset >= v6 however get version_1 because of the logic in the handler_helper. We might need a common solution for the cases that a frontend handler invokes multiple ONNX operators which has different opset versions.

tjingrant: My 2 cent: @fumihwh and I went over this issue a few days back. Some thoughts:
Firstly, @fumihwh started to call handlers of auxiliary nodes directly, c.f. how Transpose handler is used here: https://github.com/onnx/onnx-tensorflow/blob/93954860a66eab98f488da724da488a85e149166/onnx_tf/handlers/frontend/conv_mixin.py. Although I think this code can be improved as well, since here we are passing in a onnx node where the frontend handler expects a tf node. No big technical difference, but we in the future may find it less confusing looking back.

Secondly, we should modify how handlers behave to handle the complication you described. In my opinion, we should add another parameter to the `handle` method of handler called "up_to_version" which specify maximum op version handler is allow to produce. Then to create an auxiliary node, we should call that handler's `handle` function and pass in either 

1. the opset version user provide through API or CLI, 
2. or the SINCE_VERSIOIN that the op handler corresponds to. 

The latter seems simpler to implement, but the former will give user a maximally up-to-date model without breaking user specified opset constraint.

chinhuang007: @tjingrant @fumihwh  Thanks for the comments. I changed the code to call cast handler directly. It works, just a little strange to construct a ONNX node to feed the other frontend handler.

PR #307, opened 2018-10-31
onnx-tensorflow: Remove ':' while using placeholder

parth1595:
Some onnx converted model when used to convert from onnx to
 tensorflow this issue comes into the picture because the
 converted models having name data:0 and thus it will throw error
 while creating placeholder

Signed-off-by: Parth Raichura <parth.raichura@softnautics.com>


fumihwh:
@parth1595 
Thanks for contribution.
Could you provide any model that could reproduce the problem?

parth1595:
@fumihwh I have attached the onnx model here you can verify that. 
[cat_dog.onnx.zip](https://github.com/onnx/onnx-tensorflow/files/2537154/cat_dog.onnx.zip)

fumihwh:
@parth1595 
I have some questions.
First, why the onnx graph has input `data:0`?
Second, how to deal with if the graph contains `data:0`, `data:1`?

parth1595:
@fumihwh 
The onnx graph has input data:0 because it is converted from other framework which does that. So I thought we should handle this in converter. And I have added the code for the same so if data:0 or data:1 comes as name of input it will discard the character after ':'  so only data will be used in placeholder.

fumihwh:
@parth1595 
So you mean if `data:0` and `data:1` both in graph, you will handle them as same name `data`? I think it is problematic.

parth1595:
@fumihwh 

No i don't think it is problematic. Here if you notice the model the original model was not having name data:0. Before it is being converted to onnx the name of the input was data so while conversion to onnx it was appending :0. So if you consider the present case the original model can't have 2 input name as data. And thus there cannot be issue to remove that part.

chinhuang007:
@parth1595 It has been quite a few months. Please let us know if you want to continue this PR? If not, we will close it and create a new one since this seems a common issue with onnx file from tf-onnx converter.

chinhuang007:
Opened a new PR, https://github.com/onnx/onnx-tensorflow/pull/494

PR #312, opened 2018-11-05
add deprecated decorator

fumihwh:


onnx_tf/common/__init__.py
```diff
@@ -98,3 +100,51 @@ def get_output_node_names(graph_def):
...
+    UserWarning: arg_1 of func is deprecated. It will be removed in future release.
+  """
+
+  REMOVE = " It will be removed in future release."
+
+  def __call__(self, *args, **kwargs):
+    return self.deprecated_decorator(*args, **kwargs)
+
+  @staticmethod
+  def deprecated_decorator(arg=None):
```
tjingrant: why not arg={}?

onnx_tf/common/__init__.py
```diff
@@ -98,3 +100,51 @@ def get_output_node_names(graph_def):
...
+    arg = arg or {}
+
+    if inspect.isfunction(arg):
+      warnings.warn("{} is deprecated.{}".format(
+          arg.__module__ + "." + arg.__name__, Deprecated.REMOVE))
```
tjingrant: So essentially warnings will read something like:

"Size.version_1" is deprecated? If so, any chance we can make it more explicit/user-friendly?
Like op set version 1 of Size operator is deprecated.

Ignore this comment if you think it's too much work for too little utility.

onnx_tf/common/__init__.py
```diff
@@ -98,3 +100,51 @@ def get_output_node_names(graph_def):
...
+    def func():
+      pass
+
+    UserWarning: func is deprecated. It will be removed in future release.
+
+    @deprecated({"arg": "Message", "arg_1": deprecated.REMOVE})
+    def func(arg, arg_1, arg_2):
+      pass
+
+    UserWarning: arg of func is deprecated. Message
```
tjingrant: So in message, we can write something like "Use [another op] instead"?
Sounds like a useful feature we should make use of.

onnx_tf/common/__init__.py
```diff
@@ -98,3 +100,51 @@ def get_output_node_names(graph_def):
...
+    UserWarning: arg_1 of func is deprecated. It will be removed in future release.
+  """
+
+  REMOVE = " It will be removed in future release."
```
tjingrant: Maybe a bit more meaningful name?
REMOVE -> MSG_WILL_REMOVE

tjingrant:
New version looks great.

PR #325, opened 2018-11-16
Refactor size frontend handler

chinhuang007:
Refactor size frontend handler to use TensorflowNode
instead of onnx helper make_node.

### DIFF
--- a/onnx_tf/handlers/frontend/size.py
+++ b/onnx_tf/handlers/frontend/size.py
@@ -1,11 +1,11 @@
 import tensorflow as tf
 
-from onnx.helper import make_node
 from onnx_tf.common import get_unique_suffix
 from onnx_tf.handlers.frontend_handler import FrontendHandler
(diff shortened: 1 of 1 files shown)


onnx_tf/handlers/frontend/size.py
```diff
@@ -27,10 +27,13 @@ def version_1(cls, node, **kwargs):
     if not need_cast:
       return [size_node]
 
+    node.attr['DstT'] = node.attr['out_type']
```
tjingrant: I'm just curious, do we need any attributes other than 'DstT" here?

tjingrant: If not, maybe we can just create a new attr object instead of re-using the node.attr, but add a new attr to it?

chinhuang007: We only need 'DstT'. So I created a new attrs now.

PR #326, opened 2018-11-16
switch to graph freezing and allow users to specify output node names

tjingrant:


onnx_tf/converter.py
```diff
@@ -166,20 +174,47 @@ def convert(infile, outfile, convert_to, **kwargs):
...
+          input_checkpoint=latest_ckpt,
+          output_node_names=",".join(kwargs["output"]),
+          restore_op_name="",
+          filename_tensor_name="",
+          output_graph=workdir + "/frozen_model.pb",
+          clear_devices=True,
+          initializer_nodes="")
+
+      # Load back the frozen graph.
+      with open(workdir + "/frozen_model.pb", "rb") as f:
```
fumihwh: Should we remove intermediate products?

tjingrant: Yeah we probably should.

tjingrant:
@fumihwh summary of changes:
1. Allow onnx-tf to ignore purely training related ops like queue shuffling.
2. Automatically translate QueueDequeueManyV2 to an input.
3. Allow users to specify inference graph (useful for dropout removal, we should eventually automate dropout removal for inference).
4. Refactored slightly. Also addressed ur comment.

onnx_tf/common/data_type.py
```diff
@@ -55,3 +55,18 @@ def _onnx_dtype(dtype):
   else:
     raise RuntimeError("dtype should be number or str.")
   return onnx_dype
+
+
+# TODO (tjingrant) unify _onnx_dtype into any_dtype_to_onnx_dtype
+def any_dtype_to_onnx_dtype(np_dtype, tf_dtype, onnx_dtype):
```
fumihwh: Is it OK that all three are positional args?

onnx_tf/frontend.py
```diff
@@ -84,6 +100,7 @@ def tensorflow_graph_to_onnx_graph(cls,
               node_dict=dict(node_tup),
               data_type_cast_map=onnx_graph.data_type_cast_map)
         else:
+          print(node.op_type)
```
fumihwh: Dev code?

onnx_tf/common/data_type.py
tjingrant: I'm not sure what is your concern, whenever I use any_dtype_to_onnx_dtype, I always use them as keyword arguments.

PR #348, opened 2019-01-08
fix name error

tjingrant:


tjingrant:
@fumihwh do u think there's better ways than this current approach? Apparently tf doesn't allow "_" to appear as the first character of an op name.

chinhuang007 (approved):
LGTM!

PR #361, opened 2019-01-29
support batchnorm v9

tjingrant:


chinhuang007:
The is_spatial attribute is not applicable in TF. So it looks good.

PR #364, opened 2019-01-31
Fix check cli

chinhuang007:
The graph check cli is broken with the addition of
experimental ops returned from get_frontend_coverage.
Also fixed the logging not working issue because of
earlier logging level set to warning.

### DIFF
--- a/onnx_tf/opr_checker.py
+++ b/onnx_tf/opr_checker.py
@@ -2,6 +2,7 @@ import argparse
 import warnings
 import logging
 import os
+from imp import reload
 
 from google.protobuf import text_format
 from onnx import defs
@@ -13,6 +14,7 @@ from onnx_tf.common.handler_helper import get_frontend_coverage
 from onnx_tf.pb_wrapper import TensorflowNode
 
 warnings.filterwarnings('ignore')
+reload(logging)
 logging.basicConfig(level=logging.DEBUG)
 logger = logging.getLogger()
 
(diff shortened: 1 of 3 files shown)


fumihwh:
Maybe returning 3 results is not a good feature..

chinhuang007:
It is okay to return three results so gen_opset.py could put "experimental" tags in opset_version.py. We should think about making our helper methods, considered as internal APIs, more specific and stable. We probably should not freely add or remove inputs and outputs otherwise it might cause unforeseen errors, especially there is no "complete test" to verify every piece of onnx-tf code.

fumihwh:
Agree. Add or remove inputs and outputs is risky. We should not do this. In this case, dict would be better than three result.

chinhuang007:
@fumihwh Thanks for the comments. I believe you suggested to change the helper method to return a dict so we could add more content later without affecting existing code. Please review.

onnx_tf/opr_checker.py
```diff
@@ -13,6 +14,7 @@
 from onnx_tf.pb_wrapper import TensorflowNode
 
 warnings.filterwarnings('ignore')
+reload(logging)
```
fumihwh: BTW, what reload logging for?

chinhuang007: The reload clears possibly configuration by other modules and makes sure the logging setting in our code is actually used, in this case level=logging.DEBUG. In my test, the logging level in use is always the default WARNING if there is no reload.

Issue #362, opened 2019-01-29
NotImplementedError: Cast version 9 is not implemented.


davidenitti:
**Describe the bug**

NotImplementedError: Cast version 9 is not implemented.

**To Reproduce**

onnx-tf convert -t tf -i net.onnx -o net.pb

where net.onnx contains a Cast v9 operation (I suppose)

**Python, ONNX, ONNX-TF, Tensorflow version**
Python version:
3.6.7 (default, Oct 25 2018, 09:16:13) 
[GCC 5.4.0 20160609]
ONNX version:
1.4.1
ONNX-TF version:
1.2.1
Tensorflow version:
1.12.0

chinhuang007:
The fix is merged.

PR #368, opened 2019-02-02
Update Cast handlers for version 9

chinhuang007:
The Cast operator was updated in ONNX opset version 9, causning
frontend and backend conversion to fail.

This PR should fix the issue:
https://github.com/onnx/onnx-tensorflow/issues/362


onnx_tf/handlers/backend/cast.py
```diff
@@ -20,3 +20,10 @@ def version_1(cls, node, **kwargs):
   @classmethod
   def version_6(cls, node, **kwargs):
     return [cls.make_tensor_from_onnx_node(node, **kwargs)]
+
+  @classmethod
+  def version_9(cls, node, **kwargs):
+    inp = kwargs["tensor_dict"][node.inputs[0]]
+    if (node.attrs.get("to") == tf.string):
```
tjingrant: It seems like casting **from** string type would cause problem?
For instance, if we want to convert "100" to 100, this handler would err.

chinhuang007: @tjingrant good catch! Updated handler and test code to cover cast from strings. Please review.

PR #382, opened 2019-02-22
Add frontend and backend support for Where

chinhuang007:
Add frontend and backend support for Where operator which was
introduced in ONNX opset version 9.

Also filter out ConstantFill test cases since the operator
is removed from ONNX.

### DIFF
--- a/onnx_tf/handlers/frontend/select.py
+++ b/onnx_tf/handlers/frontend/select.py
@@ -1,0 +1,12 @@
+from onnx_tf.handlers.frontend_handler import FrontendHandler
+from onnx_tf.handlers.handler import onnx_op
+from onnx_tf.handlers.handler import tf_op
+
+
+@onnx_op("Where")
+@tf_op("Select")
+class Select(FrontendHandler):
+
+  @classmethod
+  def version_9(cls, node, **kwargs):
+    return cls.make_node_from_tf_node(node)
--- a/doc/support_status.md
+++ b/doc/support_status.md
@@ -37,7 +37,7 @@ ______
 |DepthToSpace|1|
 |Div|1, 6, 7|
 |Dropout|1, 6, 7|
-|DynamicSlice|1|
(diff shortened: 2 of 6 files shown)


onnx_tf/handlers/frontend/where.py
```diff
@@ -0,0 +1,13 @@
+from onnx_tf.handlers.frontend_handler import FrontendHandler
+from onnx_tf.handlers.handler import onnx_op
+from onnx_tf.handlers.handler import tf_op
+
+
+@onnx_op("Where")
+@tf_op("Select")
+
+class Where(FrontendHandler):
```
fumihwh: frontend's class name follows tf op's name.

chinhuang007: @fumihwh Good catch! Please review.

onnx_tf/handlers/frontend/select.py
```diff
@@ -0,0 +1,13 @@
+from onnx_tf.handlers.frontend_handler import FrontendHandler
+from onnx_tf.handlers.handler import onnx_op
+from onnx_tf.handlers.handler import tf_op
+
+
+@onnx_op("Where")
+@tf_op("Select")
+
```
fumihwh: Does decorator work with blank line here?

chinhuang007: Decorator seems to work with the blank line. For consistency with common code practices, I removed the blank line.

chinhuang007: @fumihwh Could you provide another review and let me know any changes needed?

fumihwh: @chinhuang007 The rest is good.

chinhuang007: @fumihwh Thanks for the review. If everything is okay, could you provide the approval so we could add this support to our master?

PR #385, opened 2019-02-28
[Back to WIP]improve strided slice frontend, implement scatter backend

tjingrant:


onnx_tf/handlers/backend/scatter.py
```diff
@@ -0,0 +1,35 @@
...
+
+@onnx_op("Scatter")
+@tf_func(tf.scatter_nd)
+class Scatter(BackendHandler):
+
+  @classmethod
+  def version_9(cls, node, **kwargs):
+    tensor_dict = kwargs["tensor_dict"]
+    axis = node.attrs.get("axis", 0)
+    assert axis == 0, "Non-zero axis is not supported in Scatter."
```
chinhuang007: This will not pass the ONNX unit test, leaving our CI always in the failed state. Shall we come up with some general approach to take care the cases where we can only provide partial support to ONNX specs? For example, using strict mode to differentiate asserts vs warnings could be an option. Just do not like the fact that we will never pass the ONNX tests :-)

tjingrant: In the past, we blacklist the particular onnx unit test we do not support.

Since the goal is to support strided slice, I don't see this restriction as a big issue at the moment. What do you think?

chinhuang007: We can put the op on the blacklist so it will not go through the ONNX unit test. That to me is not very elegant solution because we won't be testing the part that is supposed to work. The bottom line is I don't think we should merge a PR if we know it fails the ONNX unit test.

chinhuang007: If the goal of scatter is to support strided slice, not fully support the ONNX specs for scatter, I would add it to the blacklist as the short term solution.

tjingrant:
The `zeros_like` handler implemented here: https://github.com/onnx/onnx-tensorflow/pull/366 is not very useful in practice because seldom is the input tensor a constant.

winnietsang:
@tjingrant which ONNX operator would you suggest to map to tf.ZeroLike

tjingrant:
@winnietsang I've provided a more general implementation here: https://github.com/onnx/onnx-tensorflow/blob/strided_slice/onnx_tf/handlers/frontend/zeros_like.py .
However, it only supports float32 data type. Feel free to use this as a starting point to create a more general zeros_like handler. What's missing in my implementation is the ability to produce output tensors with the data type consistent with input tensor type. Since ONNX graph construction does not automatically create type information, I'm not sure how realistic it is to achieve this objective. 

That being said, if we assume the input tensor is from tensorflow graph (as opposed to something we created using TensorflowNode in frontend handler), then tensorflow probably annotates the tensor types in their NodeProto and we maybe able to use this information to perform type casting.

onnx_tf/handlers/frontend/zeros_like.py
```diff
@@ -1,14 +1,36 @@
...
 from onnx_tf.handlers.frontend_handler import FrontendHandler
+from onnx_tf.handlers.frontend.shape import Shape
 from onnx_tf.handlers.handler import onnx_op
 from onnx_tf.handlers.handler import tf_op
+from onnx_tf.pb_wrapper import TensorflowNode
 
 
-@onnx_op("Constant")
+@onnx_op("ConstantOfShape")
 @tf_op("ZerosLike")
```
chinhuang007: This handler is not getting triggered since tf.zeros_like python api produces a node with op_type='const' or 'fill', according to https://github.com/tensorflow/tensorflow/blob/6612da89516247503f03ef76e974b51a434fb52e/tensorflow/python/ops/array_ops.py#L1803. The similar message is in tensorflow release.md, 'ZerosLike and OnesLike ops treated as constants by Graph Transform Tool.'

winnietsang: We should remove the frontend handler for zero_like, but in our support document need to tell user zero_like is supported.

tjingrant: @chinhuang007 the link you gave point to the implementation of tf.zeros.

winnietsang: @tjingrant zeros_like call zeros_like_impl and zeros_like_impl call zeros

tjingrant: > ZerosLike and OnesLike ops treated as constants by Graph Transform Tool.

I fail to see how this pertain to our discussion. ZerosLike may appear to be constant for graph transform tool, but for us it is clearly not constant. Just think about tf.zeros_like whose input is a batch of images with dynamically determined batch size.

tjingrant: @winnietsang the path you pointed out is existential not universal. I can send you an email with a model with ZeroLike ops exported with tf 1.13.1 .

chinhuang007: @tjingrant The fact that the tf.zeros_like python api produces a node with op_type='const' or 'fill' is relevant in our test case as minimum. Please note test_cases.append(("test_zeros_like", **tf.zeros_like,** "zeros_like", [get_rnd([5, 5])], {})) will never trigger our handler, for the obvious reason, the op_type is NOT "ZerosLike"!

tjingrant: @chinhuang007 I agree the test case is insufficient to trigger this handler. But this is a problem of our testing methodology in this particular context, not a problem for the implementation of this handler. And this handler certainly gets triggered outside our tests on real models.

chinhuang007: @tjingrant Please send a model with ZerosLike ops exported with tf 1.13.1. We will need to find a way to write test cases for this kind of ops that do not have a matching python api. Otherwise we would never know for sure whether the handler works or not.

tjingrant: @chinhuang007 already did : )

chinhuang007: @tjingrant Just saw it :-) It would become part the graph because it is a legit C++ op. Honestly I am very interested in how it got into the graph. Regardless, the question remains, how do we test and verify our frontend handlers for these ops with no python APIs?

onnx_tf/handlers/backend/scatter.py
```diff
@@ -0,0 +1,35 @@
...
+    #       zero everywhere else) U.
+    # 2. Create dense update mask (tensor with the same shape as U,
+    #       with 1 at non-zero positions in U and 0 everywhere else) M.
+    # 3. The ONNX output tensor is therefore O = data * (1-M) + U.
+    data = tensor_dict[node.inputs[0]]
+    indices = tensor_dict[node.inputs[1]]
+    updates = tensor_dict[node.inputs[2]]
+
+    data_shape = tf.cast(tf.shape(data), indices.dtype)
+    dense_update = tf.scatter_nd(indices, updates, data_shape)
```
shawn-g-hu: Not sure if this pull request is still being looked at, but the indexing conventions for ONNX and TF differ, so in general this will fail even if axis=0.

https://github.com/onnx/onnx/blob/master/docs/Operators.md#Scatter 

vs

https://www.tensorflow.org/api_docs/python/tf/scatter_nd

chinhuang007:
This seems out-dated and replaced by https://github.com/onnx/onnx-tensorflow/pull/511. Please reopen if we still need it.

PR #387, opened 2019-03-01
Test fix for AttributeError: 'function' object has no attribute 'args' 

tjingrant:


tjingrant:
Fix credit to @fumihwh .

PR #388, opened 2019-03-01
Fix CI error for older versions of onnx and TF

chinhuang007:
Fix a CI error for older versions of onnx and TF in the test code
for cast.


winnietsang:
LGTM! The string support for Cast is added in opset 9, so the test for that should only be tested for version >= 9.

PR #396, opened 2019-03-17
update travis to use docker

tjingrant:


tjingrant:
@winnietsang @chinhuang007 @fumihwh , after learning and struggling with Docker for a good amount of time, I think this patch is ready. So once we tag the current tip of master as "onnx1.3.0-tf1.13.1" (which is also the docker image name winnie published), we will be able to have a green travis badge corresponding to the onnx-tf tested with onnx1.3.0 and tf1.13.1.

Following this patch I will,
- Tag and post a release.
- Get the green travis badge url for the tagged release.
- Update Pypi.
- Update the README in a separate patch to include our releases and their build status.

tjingrant:
@chinhuang007 I'm not sure either, that's why I left it there. @winnietsang can you comment on whether we should retain these RUN commands?

winnietsang:
@tjingrant I don't think those lines are needed anymore, we should just remove them.

PR #399, opened 2019-03-25
fix onnx(pytorch) to tf pb upsample bug

RoGoSo:
for the onnx model that generate from pytorch  will get "Upsample without nearest or bilinear". So I change some in the code and fix that.


silfverstrom:
I missed that this PR tries to solve the same problem as my PR.
https://github.com/onnx/onnx-tensorflow/pull/416

However, There is a bug in this PR. It will always cause the interpolation to use bilinear.
We could update this PR or merge the one i proposed above.

onnx_tf/handlers/backend/upsample.py
```diff
@@ -34,7 +34,7 @@ def version_7(cls, node, **kwargs):
     new_weight = np.floor(x_shape[3] * scales[3])
 
     mode = attrs.get("mode", "nearest")
-    if mode.lower() == "bilinear":
+    if mode.lower() == "bilinear" or "linear":
```
silfverstrom: This is not valid syntax. It would always return true, and thus always use bilinear interpolation.

chinhuang007:
The issue should be fixed by PR #416. Feel free to reopen it not the case.

PR #417, opened 2019-05-10
Fix slice backend handler

chinhuang007:
Fix the slice backend handler. The issue is due to a recent
change for opset 10 broke the opset 9 support.


chinhuang007:
Additional work has to be added to support steps. I am going to add the changes to this PR. So please no need to review yet.

chinhuang007:
@tjingrant @fumihwh Added support for variations of steps input. Please review.

chinhuang007:
Thanks for the review. I just added a test case with non-default steps.

PR #452, opened 2019-07-03
GRU support with linear_before_reset

iiSeymour:
Similar to #227 but without adding the backend, use the [CudnnCompatibleGRUCell](https://www.tensorflow.org/api_docs/python/tf/contrib/cudnn_rnn/CudnnCompatibleGRUCell) implementation in TensorFlow instead.

### DIFF
--- a/onnx_tf/handlers/backend/gru.py
+++ b/onnx_tf/handlers/backend/gru.py
@@ -18,9 +18,6 @@ class GRU(RNNMixin, BackendHandler):
     num_directions = 2 if direction == "bidirectional" else 1
     if "clip" in node.attrs:
       exception.OP_UNSUPPORTED_EXCEPT("GRU with clip", "Tensorflow")
-    if node.attrs.get("linear_before_reset", 0):
-      exception.OP_UNSUPPORTED_EXCEPT("GRU with linear_before_reset",
-                                      "Tensorflow")
     if "activations" in node.attrs:
       activations = list(map(lambda x: x.lower(), node.attrs["activations"]))
       if activations[0] != "sigmoid":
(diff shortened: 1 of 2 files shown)


chinhuang007:
Looks like https://www.tensorflow.org/api_docs/python/tf/contrib/cudnn_rnn/CudnnCompatibleGRUCell is deprecated. In general, we should avoid using APIs in contrib. We would need to replace it soon anyway so could you find alternative ways?

iiSeymour:
It looks like the new [GRU](https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/keras/layers/GRU) layer is CuDNN compatible and supports `reset_after` so I think it should just be a case of swapping `tf.contrib.cudnn_rnn.CudnnCompatibleGRUCell` for `tf.keras.layers.GRU`. I will have a proper look tomorrow.

iiSeymour:
Needs #460 fixing first.

dianyo:
Hi, does this feature still on progress? I met this problem when I transfer my ONNX model. I'd like to fix this, if there's no implement yet

alamnasim:
> Hi, does this feature still on progress? I met this problem when I transfer my ONNX model. I'd like to fix this, if there's no implement yet

Any update on this? I am also facing the same issue while converting ONNX to tensorflow.

iiSeymour:
The patch I submitted worked for me at time submission but I don't have time to (re)evaluate currently @alamnasim @dianyo give it a try.

tylerweitzman:
Any update on this?

TingfengTang:
@iiSeymour, when I use your gru-reset-after branch,I found the value of the argument 'name' in function _custom_getter() which defined in gru.py is 'GRU_9c8f1236/rnn/multi_rnn_cell/cell_0/cudnn_compatible_gru_cell/candidate/hidden_projection/bias'. Can you tell me how is the argement name generated, I have not found it in the context, thanks.

PR #456, opened 2019-07-05
Add max_unpool operation to backend

sdmonov:
MaxUnpool is not supported by tensorflow by default. Refer to https://github.com/tensorflow/tensorflow/issues/2169 for more information. The current solution uses proposed code from the above issue with modifications to support for padding and strides.


onnx_tf/handlers/backend/unpool_mixin.py
```diff
@@ -0,0 +1,76 @@
...
+                                          shape=[input_shape[0], 1, 1, 1])
+      b = tf.ones_like(ind) * batch_range
+      b1 = tf.reshape(b, [flat_input_size, 1])
+      ind_ = tf.reshape(ind, [flat_input_size, 1])
+      ind_ = tf.concat([b1, ind_], 1)
+
+      ret = tf.scatter_nd(ind_, pool_, shape=tf.cast(flat_output_shape, tf.int64))
+      ret = tf.reshape(ret, output_shape)
+
+      set_input_shape = pool.get_shape()
```
chinhuang007: set_input_shape is not used anywhere, maybe remove it? Please use pylint to do a quick check as recommended in https://github.com/onnx/onnx-tensorflow. I wouldn't worry about missing docs but do want to keep the code clean.

sdmonov: set_input_shape is left over from old code. Will remove it. There is one more issue with the code that I'm working on right now. It's related to how to consider output_shape in the calculations. It is not strictly specified in the ONNX ops documentation but from the examples given (and the unit tests) looks like the output shape specifies just padding of the output and shouldn't be used during the unpooling. I'm working on this right now and should have new changes soon.

winnietsang:
@sdmonov please update onnx-tensorflow/doc/support_status.md by running 
 onnx-tensorflow/onnx_tf/gen_doc.py

sdmonov:
@winnietsang done. Thanks for pointing it out.

PR #457, opened 2019-07-05
Update CI to use proper python versions

chinhuang007:
Made a few minor updates
1. add python version check to run different pythons in docker
2. change base OS to xenial
3. remove script for release tag so we simply run CI for master

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 os:
   - linux
 #  - osx
-dist: trusty
+dist: xenial
 sudo: required
 language: python
 services:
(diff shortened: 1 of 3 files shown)


tjingrant:
Hi Chin, seeing this change:
> remove script for release tag so we simply run CI for master

I wonder how we should test release branch against onnx/tf release version?

chinhuang007:
I believe you are talking about the test of the latest onn-tf release branch against newer onnx/tf release versions. This kind of tests should be handled separately, not during each PR push or merge. The release branch test is only needed when onnx or tf has an official new release, which happens every few months. We could add it back if we can set a trigger on ONNX an tf new releases in travis, or put it as periodic tests if we can't set such triggers. Either way I don't see the connection between a PR push/merge with release tests, therefore the remove for now.

tjingrant:
@chinhuang007 thanks for the explaination, in this case I think it's probably okay to remove the conditional about release tests.

chinhuang007:
Thanks for the review.

PR #466, opened 2019-07-16
Fix test error for ConvTranspose

chinhuang007:
TF currently doesn't support dilations != 1 in convxd_transpose.
Therefore comment out the onnx test case where dilations set to 2.


chinhuang007:
@tjingrant This PR removes the last error in CI so the status would be green.

Issue #465, opened 2019-07-15
lp_normalization.py implementation is not right


wutiantong:
Currently, the implementation is equal to `tf.norm()`, which is ReduceL2 actually.

Totally speaking, lp_normalization should be `tf.div(X, tf.norm(X))`.

chinhuang007:
Thanks for pointing it out. Will look into a fix.

chinhuang007:
The fix is merged. Please go ahead and verify. Reopen if the problem is not addressed.

PR #474, opened 2019-07-30
Fix LpNormalization handler

chinhuang007:
Fix LpNormalization handler to return the expected output as
reported in issue https://github.com/onnx/onnx-tensorflow/issues/465

Also run yapf to ensure coding standard format.


chinhuang007:
@tjingrant @fumihwh Please help review.

PR #480, opened 2019-08-06
Add opet 11 support for DepthToSpace

chinhuang007:
Add opet 11 support for DepthToSpace with the introduction
of mode "CRD"


chinhuang007:
@tjingrant @fumihwh Please take a look at this native computation for the new mode. Feel free to suggest alternative solutions.

onnx_tf/handlers/backend/depth_to_space.py
```diff
@@ -27,3 +27,38 @@ def version_1(cls, node, **kwargs):
...
+    x_rank = len(x.get_shape())
+    storage_format, compute_format = get_data_format(x_rank)
+    attrs = copy.deepcopy(node.attrs)
+    attrs["data_format"] = storage_format
+    mode = attrs.get("mode", "DCR")
+    block_size = attrs.get("blocksize")
+
+    if mode == "CRD":
+      # need native computation
+      bsize = tf.constant(block_size)
```
tjingrant: It looks like you don't need `tf.constant` here?
It seems like you can just use python variables throughout.

chinhuang007:
Thanks for the review and comments. Using python is easier!

PR #487, opened 2019-08-22
Add SparseTensor support for Constant version 11

chinhuang007:
The SparseTensor data type and as an attribute for Constant
is introduced in ONNX opset 11. This PR adds the support for
the new data type in constant handler.


chinhuang007:
Hi Tian, thanks for the comments. The code should look much cleaner now.

PR #488, opened 2019-08-25
Add Opset v10 support for MaxPool

sdmonov:
Version 10 support includes dilations and ceil_mode.

This is new implementaion of dilated pooling by calculating and extracting (using tf.gather_nd) the values from the input Tensor that match the kernel, strides and dilations. It utilizes tf.nn.max_pool2d or tf.nn.max_pool_with_argmax to extract the final max values/indices.
Also the code will utilize tf.nn.dilation2d when possible.


chinhuang007:
Please take a look at strict argument, https://github.com/onnx/onnx-tensorflow/blob/master/doc/CLI.md and see if it has any impact to the new version.

sdmonov:
@chinhuang007 the strict argument will not have influence on the v10 code as it will always generate semantically equivalent tensorflow model to the orignal model

chinhuang007:
@tjingrant Please take a look since you have much better knowledge about the strict mode. Should we keep it and document that it has no effect in v10? or Merge pool() and pool_v10() which is time-consuming and might cause regression?

onnx_tf/handlers/backend/dilated_maxpooling.py
```diff
@@ -0,0 +1,430 @@
...
+        input_shape = _tf_shape(self.input)
+        in_spatial_shape = input_shape[1:self.spatial_size+1]
+
+        channel_num = input_shape[self.spatial_size+1]
+        gather_ind = tf.range(channel_num, dtype=tf.int64)
+        gather_ind = tf.expand_dims(gather_ind, 1)
+
+        self.output_shape = np.zeros([self.spatial_size + 2])
+        self.output_shape[0] = input_shape[0]
+        for dim in range(self.spatial_size - 1, -1, -1):
```
tjingrant: A bit more inline comments in this function (_reduce_dilations) will be greatly appreciated.
I'm having a hard time understanding the detailed implementation here.

onnx_tf/handlers/backend/dilated_maxpooling.py
```diff
@@ -0,0 +1,430 @@
...
+        gather_ind = tf.tile(gather_ind, [input_shape[0]] + [1] *
+                             (self.spatial_size + 1))
+
+        # extract the selected values from the input
+        output = tf.gather_nd(self.input, gather_ind, batch_dims=1)
+        output = tf.reshape(output, self.output_shape)
+
+        return output
+
+    def _calc_pads_same(self, in_spatial_shape):
```
tjingrant: Can we unify this with calc_pads_same from pooling helper?

onnx_tf/handlers/backend/dilated_maxpooling.py
```diff
@@ -0,0 +1,430 @@
...
+        padded_width = self.input_shape[2]
+        num_channels = self.input_shape[3]
+
+        # mod_floor op is not implemented on GPU
+        # implement it using: a % b = a - (a // b) * b
+
+        # ind_ = ind // num_channels
+        # ind_channel = ind % num_channels
+
+        ind_ = ind // num_channels
```
tjingrant: Again, can we have a bit more descriptive variable names.

onnx_tf/handlers/backend/dilated_maxpooling.py
tjingrant: OK, second look I got what you are doing here; but it would still be nice to explain that you are building a list of N-D index to extract values to reduce-pool. Also please add in comment some high-level design here, like you are calculating the output index and transform it back to input index by reusing the _calc_input_ind function. As it's a nice and simple trick worth explaining.

onnx_tf/handlers/backend/onehot.py
```diff
@@ -38,7 +38,7 @@ def version_9(cls, node, **kwargs):
     return [
         cls.make_tensor_from_onnx_node(
             node,
-            inputs=[indices, depth[0], on_value, off_value],
+            inputs=[indices, depth, on_value, off_value],
```
tjingrant: The change looks right to me. But just curious, any reason for this change? Did you get any trouble form using the code before this change?

test/backend/test_onnx_backend.py
```diff
@@ -87,6 +87,10 @@ def skip_not_implemented_ops_test(test):
 # skip resize downsample with mode=linear
 backend_test.exclude(r'test_resize_downsample_linear[a-z,_]*')
 
+# range is using loop in the unit test but
+# loop is not supported in tf-onnx converter
```
tjingrant: Do you mean tf-onnx or onnx-tf?

onnx_tf/handlers/backend/pool_mixin.py
```diff
@@ -115,6 +115,64 @@ def pool(cls, node, input_dict, pool_func, pooling_type, strict=True):
...
+    if need_trans:
+      compute_format = "N" + storage_format[2:] + "C"
+      x = tf.transpose(x, perm=get_perm_from_formats(storage_format,
+                                                     compute_format))
+
+    dp = DilatedPooling(input=x, kernel_shape=kernel_shape, strides=strides,
+                        dilations=dilations, padding=pads, ceil_mode=ceil_mode)
+
+    result = []
+    if pooling_type == "MAX":
```
tjingrant: I'm curious, any chance we can simplify this cross product here? pooling type x need_trans?
Something like this:
preprocess_op = tf.transpose... if need_trans else tf.identity...
pool_op = dp. dilated_maxpool if pooling_type == "MAX" else dilated_maxpool_with_argmax

unpacking results may be tricky, please let me know if you think its improvable.

sdmonov:
@tjingrant Thank you very much for your review. I'm working on addressing your comments and requests. I'll inform you when I have new version for review. Thanks!

onnx_tf/handlers/backend/dilated_maxpooling.py
```diff
@@ -0,0 +1,430 @@
...
+        self.input_shape = self.orig_input_shape
+
+        if pooling_type == "MAX":
+            self.padding_constant = -inf
+        else:
+            self.padding_constant = 0
+
+    def _calc_input_ind(self, ind, k, d, s):
+        """
+            This function maps index from the output of _reduce_dilations
```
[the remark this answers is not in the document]
sdmonov: Added comments. Please go through them and tell me if it is better now.

onnx_tf/handlers/backend/dilated_maxpooling.py
sdmonov: Added comments in multiple places and tried to make the code mode readable.

onnx_tf/handlers/backend/onehot.py
sdmonov: this was onehot.py was some issue with rebasing before. I didn't work on it. Now the commits should be in order and onehot.py is not part of this PR.

test/backend/test_onnx_backend.py
sdmonov: same as for onehot.py. this change is not supposed to be part of this PR. fixed now.

onnx_tf/handlers/backend/pool_mixin.py
sdmonov: good idea. had to use lambda functions.

sdmonov:
@tjingrant made multiple changes to the code. please review it again. Thank you!

chinhuang007:
@sdmonov Please resolve the branch conflicts with master.

chinhuang007:
@sdmonov I merge the PR. But there are two minor places to be cleaned up. Please check the LGTM analysis alerts (basically extra and unused imports) and fix them in the average pool PR. Thanks!

sdmonov:
@chinhuang007 sure I saw the LGTM alerts. I think I can merge them with the average pool PR.
Thanks!

theotheo:
@sdmonov I'm sorry to trouble you. But I think you could be good advice about this issue https://github.com/onnx/onnx-tensorflow/issues/595
Is it possible to substitute `dilation2d` with something supported by Tensorflow Lite?

Any help would be appreciated!

sdmonov:
@theotheo can you take a look at #613. It should solve the issue. Thanks!

PR #494, opened 2019-08-30
Fix placeholder name containing ":"

chinhuang007:
Fix placeholder name containing ":", replace ":..." with
"_" + unique_suffix so the new name should not collide with
other node names in the graph.


onnx_tf/backend.py
```diff
@@ -106,9 +107,13 @@ def _onnx_graph_to_tensorflow_rep(cls, graph_def, opset, strict):
         shape = list(
             d.dim_value if (d.dim_value > 0 and d.dim_param == "") else None
             for d in value_info.type.tensor_type.shape.dim)
+
+        value_info_name = value_info.name if (value_info.name.find(
```
tjingrant: 1. I think string containment can be done in a more pythonic way 
```
if (value_info.name.find(":") == -1)
```
can be rewritten as:
```
if ":" in value_info.name
```
2. The targeted scenario seems to be a bit too specific; although the method as shown below works for cases where the name contains more than one ":", I feel like it's not really working in a way as one would expect. Can we just perform string replacement? ":" -> "_" so that no matter how many ":"s there are, we can always take care of the situation in an elegant way?

chinhuang007: Thanks for the review! I changed to use the more pythonic way. As for replacing ":" with "_", it won't work if there are node names "test:0" and "test_0" in the model. The current solution of course doesn't preserve the string after ":" but gets a unique and valid node name. So the question is do we want to develop a sophisticated name conversion mechanism to handle a special case that the onnx model is converted from TF using tf-onnx? And how important it is to be able to convert back and forth between node name with ":" (could be many) and one without?

chinhuang007: The related issue in tf-onnx is https://github.com/onnx/tensorflow-onnx/issues/539

tjingrant: Hi Chin, I think we should preserve whatever comes after the ":". Because the users need to supply our backend (or tensorflow) with the exact name of the output tensor to retrieve its output value. Imagine, in the onnx model, the output node has two output tensors with name "op:0" and "op:1", looking at the converted tensorflow model, the user will be completely clueless about the correspondence between {"op:0", "op:1"} and {"op1234", "op4321"} and unable to easily determine which output tensor to retrieve as the prediction. On the other hand, users have much more clue if the mangled names are as such {"op_0_1234", "op_1_4321"}.

Also, maybe we can throw in an extra check:
https://github.com/tensorflow/tensorflow/blob/2b7d3c9a90f17ce8ab1ebe535c08af4abd9836e1/tensorflow/python/framework/ops.py#L1565

We can verify whether the onnx name is legal in tensorflow, and throw a more user friendly message ourselves, instead of waiting for tensorflow to panic.

chinhuang007: Okay. In that case I would suggest xxx_yyy_zzz_UniqueString for xxx:yyy:zzz.

chinhuang007: The fix is checked in, replace ":" with "_tf_" and append "_" + unique_suffix. Please review. 

By the way, the code is only affecting the value_info in the graph inputs, making sure valid names for the placeholders we create on the fly. It is not to check and replace all node names in the graph. Therefore the output tensor name concern seems nonexistent.

chinhuang007: I would suggest open a new issue/PR to check and verify all node names in the onnx graph are valid in Tensorflow, and leave this PR to solve a very specific issue with ":" in value_info in the graph inputs which is produced by tf-onnx converter.

Issue #327, opened 2018-11-19
Bidir LSTM `activations` attribute is optional


samgd:
**Describe the bug**

The ONNX [LSTM operator](https://github.com/onnx/onnx/blob/master/docs/Operators.md#attributes-26) states that the `activations` attribute is optional. The [LSTM backend implemenation](https://github.com/tengyifei/onnx-tensorflow/blob/dfc6f01fcb6aa2a967de4e5edc9371818af1fc1a/onnx_tf/handlers/backend/lstm.py#L13) relies on this property being present when using a bidirectional LSTM. 

The implementation initialises `tf_activations = [tf.nn.tanh]` and only appends an additional activation function `if "activations" in node.attrs:"` and `if num_directions == 2:` [here](https://github.com/tengyifei/onnx-tensorflow/blob/dfc6f01fcb6aa2a967de4e5edc9371818af1fc1a/onnx_tf/handlers/backend/lstm.py#L131):

```python
    tf_activations = [tf.nn.tanh]
    if "activations" in node.attrs:
      activations = list(map(lambda x: x.lower(), node.attrs["activations"]))
      activation_alpha = node.attrs.get("activation_alpha", [None] * 6)
      activation_beta = node.attrs.get("activation_beta", [None] * 6)
      tf_activations = [
          cls.rnn_get_activation(activations[1], activation_alpha[1],
                                 activation_beta[1])
      ]
      if num_directions == 2:
        tf_activations.append(
            cls.rnn_get_activation(activations[4], activation_alpha[4],
                                   activation_beta[4]))
```

`tf_activations` is passed to the `rnn` method [here](https://github.com/tengyifei/onnx-tensorflow/blob/dfc6f01fcb6aa2a967de4e5edc9371818af1fc1a/onnx_tf/handlers/backend/lstm.py#L179):

```python
      outputs, states = cls.rnn(x, tf.nn.rnn_cell.LSTMCell, cell_kwargs,
                                rnn_kwargs, tf_activations, direction)
```

The `rnn` method relies on there being a second `activations` value present `if direction == "bidirectional":`, where `activations` is the `tf_activations` from above [here](https://github.com/tengyifei/onnx-tensorflow/blob/master/onnx_tf/handlers/backend/rnn_mixin.py#L33):

```python
    if direction == "bidirectional":
      cell_kwargs["activation"] = activations[1]
      rnn_cell_bw = [cell_class(**cell_kwargs)]
[...]

PR #497, opened 2019-09-02
fix BLSTM activations error

azraelkuan:
[the problem ](https://github.com/onnx/onnx-tensorflow/issues/327) has been fixed.


onnx_tf/handlers/backend/lstm.py
```diff
@@ -126,8 +126,11 @@ def _common(cls, node, **kwargs):
 
     if "clip" in node.attrs:
       cell_kwargs["cell_clip"] = node.attrs["clip"]
-
+  
     tf_activations = [tf.nn.tanh]
+    if num_directions == 2:
```
chinhuang007: I believe the issue https://github.com/onnx/onnx-tensorflow/issues/327 is concerning when the optional attribute "activations" is not present. If the attribute is there, the current code should be good. Maybe move the append to if "activations" in node.attrs:.... else: if num_directions == 2: tf_activations.append(tf.nn.tanh)

azraelkuan: see the new commit, sorry for late relpy

tjingrant:
@chinhuang007 please help review my edits to this patch.

PR #507, opened 2019-09-27
Add support for QLinearConv

chinhuang007:
Add support for QLinearConv which is introduced in ONNX
opset 10.

### DIFF
--- a/onnx_tf/handlers/backend/q_linear_conv.py
+++ b/onnx_tf/handlers/backend/q_linear_conv.py
@@ -1,0 +1,73 @@
+import tensorflow as tf
+
+from onnx_tf.handlers.backend_handler import BackendHandler
+from onnx_tf.handlers.handler import onnx_op
+from .conv_mixin import ConvMixin
+
+
+@onnx_op("QLinearConv")
+class QLinearConv(ConvMixin, BackendHandler):
+
+  @classmethod
+  def _dequantize_tensor(cls, base, zero_point, scale):
+    # Do computation in float32
+    base = tf.cast(base, tf.float32)
+    zero_point = tf.cast(zero_point, tf.float32)
+    return (base - zero_point) * scale
+
+  @classmethod
+  def _dequantize_w(cls, base, zero_point, scale):
+    tensor_list = [
+        cls._dequantize_tensor(base[i][j], zero_point[j], scale[j])
+        for i in range(base.shape.as_list()[0])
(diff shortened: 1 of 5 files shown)


onnx_tf/handlers/backend/q_linear_conv.py
```diff
@@ -0,0 +1,70 @@
...
+    for i in [7, 6, 5, 4, 2, 1]:
+      node.inputs.remove(node.inputs[i])
+
+    # Use common conv handling
+    conv_node = cls.conv(node, kwargs["tensor_dict"])
+
+    # Process output
+    y = tf.round(conv_node / y_scale) + y_zero_point
+
+    return tf.cast(y, tf.int32)
```
tjingrant: Why int32? It seems like spec says output is either uint8 or int8.

onnx_tf/handlers/backend/q_linear_conv.py
```diff
@@ -0,0 +1,70 @@
+import tensorflow as tf
+
+from onnx_tf.handlers.backend_handler import BackendHandler
+from onnx_tf.handlers.handler import onnx_op
+from .conv_mixin import ConvMixin
+
+
+@onnx_op("QLinearConv")
+class Conv(ConvMixin, BackendHandler):
```
tjingrant: Forgot to change name?

onnx_tf/handlers/backend/q_linear_conv.py
```diff
@@ -0,0 +1,70 @@
...
+    elif len(w_scale.shape) > 1:
+      raise ValueError("Unsupported scale: {}".format(w_scale))
+
+    # Dequantize variables to float32
+    x = cls._dequantize_tensor(x, x_zero_point, x_scale)
+    w = cls._dequantize_w(w, w_zero_point, w_scale)
+    y_zero_point = tf.cast(y_zero_point, tf.float32)
+
+    tensor_dict[node.inputs[0]] = x
+    tensor_dict[node.inputs[3]] = w
```
tjingrant: Just to make sure: so is bias automatically taken care of by the regular conv logic?

chinhuang007:
@tjingrant Thanks for spotting the errors. Made the following changes:
1. class name
2. output dtype is the input dtype, not tf.int32

The bias is handled generally in conv_mixin.py, starting line 238. I don't think it would be any different in this quantized case.

PR #511, opened 2019-10-04
Support scatter, gather

tjingrant:
I also tried to support Scatter/Scatter Elements but it seems like it is not the common Scatter operation as it seems to support scattering only on one axis. In other words, the Scatter/Scatter Elements from ONNX/PyTorch seems to be in between a real sparse operation and a dense operation as sparsity is limited to occur only in one dimension; i imagine this design is due to efficiency concerns. Supporting it in tensorflow is possible but would likely trash performance (tensorflow has the true scatter operation, which is a sparse operation where sparsity can occur on any dimensions). It feels like supporting this very specific version of ONNX Scatter/Gather elements is a great deal of efforts for no clear gain. 

Please let me know your thoughts.


tjingrant:
@chinhuang007 ScatterElements are added. 
Please let me know if the comments are sufficient. Comments are written for the high-level idea, but implementation slightly differs in ways that makes coding easier.

chinhuang007:
@tjingrant The comments are very good. Just wonder if this PR has coverage for Scatter which is introduced in opset 9 and deprecated in opset 11 in favor of ScatterElements? Related question, shell we have a handler for Scatter for opset 9-10?

tjingrant:
Hi @chinhuang007 added Scatter op for opset=9. Please let me know if there's any additional comments?

onnx_tf/handlers/backend/gather_nd.py
```diff
@@ -0,0 +1,15 @@
...
+
+@onnx_op("GatherND")
+class GatherND(BackendHandler):
+
+  @classmethod
+  def version_11(cls, node, **kwargs):
+    params = kwargs["tensor_dict"][node.inputs[0]]
+    indices = kwargs["tensor_dict"][node.inputs[1]]
+
+    return [tf.gather_nd(params, indices)]
```
chinhuang007: This seems a direct mapping to a tf_func. Wonder if we should use the @tf_func(tf.gather_nd) approach?

test/backend/test_node.py
```diff
@@ -1237,6 +1255,75 @@ def test_where(self):
     output = run_node(node_def, [c, x, y])
     np.testing.assert_almost_equal(output["Z"], np.where(c, x, y))
 
+  def test_scatter_nd(self):
```
chinhuang007: Please filter out the new tests so the target CI, currently for opset 10, would pass.

chinhuang007:
@tjingrant Thanks for the changes. Everything looks good. But there seems some conflicts with master in test_node.py. Please take a look.

tjingrant:
@chinhuang007 resolved.

PR #512, opened 2019-10-12
Add support for TfIdfVectorizer

chinhuang007:
Add support for TfIdfVectorizer


onnx_tf/handlers/backend/tfidf_vectorizer.py
```diff
@@ -0,0 +1,123 @@
...
+    if n == 1:
+      return cls._calc_ngram_skip(x, pool, n)
+
+    # Loop through maximum allowable skip count and sum up the results
+    result = tf.zeros([int(len(pool) / n)], dtype=tf.int32)
+    max_allowable_skip = np.minimum(max_skip,
+                                    int((int(x.shape[0]) - 1) / (n - 1) - 1))
+    for skip in range(max_allowable_skip + 1):
+      # For each skip calculate the ngram counts
+      result += cls._calc_ngram_skip(x, pool, n, skip)
```
tjingrant: Seems like it's reasonable to put definition of `_calc_ngram_skip` before `_calc_ngram`.

onnx_tf/handlers/backend/tfidf_vectorizer.py
```diff
@@ -0,0 +1,123 @@
...
+    return ngrams
+
+  @classmethod
+  def _calc_ngram(cls, x, pool, n, max_skip):
+    # This method calculates ngram counts for a specific n and
+    # all allowable skips
+
+    # For 1gram, skip is not in use
+    if n == 1:
+      return cls._calc_ngram_skip(x, pool, n)
```
tjingrant: I think it may be more appropriate to have a special case in `_prepare_ngrams` for better abstraction. Because I think you actually need this special case for this whole algorithm to work correctly, which implies that `_prepare_ngrams`'s abstraction breaks when n=1 and skip is not zero, meaning `_prepare_ngrams` is not really self-contained.

onnx_tf/handlers/backend/tfidf_vectorizer.py
```diff
@@ -0,0 +1,123 @@
...
+    # Make input as an array of ngrams
+    new_x = cls._prepare_ngrams(x, n, skip)
+
+    # Loop through the ngram targets in the pool
+    for i in range(len(pool)):
+      ngram_count = tf.map_fn(lambda in_x: tf.where(tf.reduce_all(tf.equal(in_x, tf.constant(pool[i], dtype=new_x.dtype))), tf.constant([1]), tf.constant([0])), new_x, dtype=tf.int32)
+      ngram_count = tf.math.count_nonzero(ngram_count, dtype=tf.int32)
+      ngram_count = tf.reshape(ngram_count, [1])
+      # Building up the result
+      if i == 0:
```
tjingrant: Can we store intermediate results in a tensor list and do a single concatenation in the end? Feels like this can potentially reduce the number of concat nodes in our imported graph.

onnx_tf/handlers/backend/tfidf_vectorizer.py
```diff
@@ -0,0 +1,123 @@
...
+          result = cls._calc_ngram(input_t, process_pool, gram_len,
+                                   max_skip_count)
+          idx = tf.constant(idx, shape=[len(idx), 1])
+          result_ngram = result_ngram + tf.scatter_nd(idx, result, [size])
+      return result_ngram
+
+    # The input can be either 1d or 2d. Need to loop through
+    # each element for 2d inputs
+    n = len(input_tensor.shape)
+    final_out = [
```
tjingrant: I appears to me that you have `process_ngram` organized as a function because you want the algorithm to work on two scenarios (1d, 2d). It may help with the clarity of the algorithm to have slightly more complicated pre/post-processing steps between which inputs are guaranteed to be 2d. This way, we reduce the complexity of the main body of the algorithm.

onnx_tf/handlers/backend/tfidf_vectorizer.py
```diff
@@ -0,0 +1,123 @@
...
+
+  @classmethod
+  def _prepare_ngrams(cls, x, n, skip):
+    # This method transform input into n-grams for a specific skip
+    # input x is a 1D tensor
+    # ex1: x=[1,2,3,4] n=1 skip=0 output=[[1],[2],[3],[4]]
+    # ex2: x=[1,2,3,4] n=2 skip=0 output=[[1,2],[2,3],[3,4]]
+    # ex3: x=[1,2,3,4] n=2 skip=1 output=[[1,3],[2,4]]
+    count = x.shape[0] - n + 1 - skip
+    mul = skip + 1
```
tjingrant: Variable names could be more intuitive, can we have `count` -> `num_ngrams` and i don't quite get what this `mul` mean.

onnx_tf/handlers/backend/tfidf_vectorizer.py
chinhuang007: Sure, method moved.

onnx_tf/handlers/backend/tfidf_vectorizer.py
chinhuang007: Please check test_node.py, # test IDF mode with weights, max_skip=5, output contains 1-grams and 2-grams. It should invoke _prepare_ngrams with n=1 and skip=5. Maybe I missed something?

chinhuang007: I read the spec one more time and don't see "For 1gram, skip is not in use". However the ONNX examples seem to indicate so. As seen in tf_batch_uniandbigrams_skip5 and tf_uniandbigrams_skip5, the 1-gram results [0, 3, 0, 0] and [0, 3, 1, 0] are not the accumulated counts from multiple skips. Therefore the special handling for 1gram is based on examples.

chinhuang007:
@tjingrant Thanks for the comments! All concerns should be addressed by changing code or adding documentation. Please review.

onnx_tf/handlers/backend/tfidf_vectorizer.py
chinhuang007: A tensor_list is used to minimize the number of concat nodes.

onnx_tf/handlers/backend/tfidf_vectorizer.py
chinhuang007: Added comments to process_ngram to indicate general purpose of processing a row of inputs.

PR #520, opened 2019-11-03
Scan operator

sdmonov:
This is support for scan operator. It implements version 8, 9 and 11 of the operator. The unit tests will skip on v11 because they rely on split op and it does not have v11 yet. The code can be easily enabled when v11 of split is ready.


sdmonov:
@winnietsang can you please review.
Thanks!

onnx_tf/backend.py
```diff
@@ -254,6 +254,27 @@ def _get_handlers(cls, opset):
   def supports_device(cls, device):
     return common_supports_device(device)
 
+  @classmethod
+  def onnx_graph_to_tensorflow_ops(cls, graph_def, input_values,
+                                   opset=None, strict=True):
```
chinhuang007: Please provide basic description for the method, params, and return.

onnx_tf/handlers/backend/scan_mixin.py
```diff
@@ -0,0 +1,188 @@
...
+from onnx.helper import make_opsetid
+import onnx_tf
+from onnx_tf.common import data_type
+
+
+class ScanMixin(object):
+
+    @classmethod
+    def scan(cls, node, input_dict, strict):
+        current_opset = [make_opsetid(cls.DOMAIN, cls.SINCE_VERSION)]
```
winnietsang: This current_opset is the opset version for the whole "body"(graph), it should be using the cls.VERSION. cls.SINCE_VERSION should only be use to determine the version of Scan handler it is going to use.

onnx_tf/handlers/backend/scan_mixin.py
```diff
@@ -0,0 +1,188 @@
...
+            outputs = [tensor_dict[output.name] for output in body.output]
+            return outputs
+
+        scan_input_axes = node.attrs.get("scan_input_axes",
+                                         [0] * num_scan_inputs)
+        # Version 8 of the operator
+        scan_input_directions = node.attrs.get("directions", None)
+        if scan_input_directions is None:
+            scan_input_directions = node.attrs.get("scan_input_directions",
+                                                   [0] * num_scan_inputs)
```
chinhuang007: Could be combined into one line to cover either version 8 or greater than 8.

onnx_tf/backend.py
sdmonov: Done!

onnx_tf/handlers/backend/scan_mixin.py
sdmonov: fixed

PR #522, opened 2019-11-05
average_pool op v10 and v11

sdmonov:
Implementation of average pooling v10 and v11


sdmonov:
@tjingrant can you please review. Thanks!

onnx_tf/handlers/backend/dilated_pooling.py
```diff
@@ -613,3 +621,18 @@ def dilated_maxpool(self, force_custom_impl=False):
...
+
+    def is_supported(self):
+        """
+            Function to check if the current set of arguments are
+            supported for average pool
+        """
+        if self.pooling_type.startswith("MAX"):
+            return True
+        else:
+            return self.count_include_pad or ((
```
tjingrant: Can we break this long logical expression down a little bit for readability?

sdmonov: @tjingrant thanks for the comment. I split this expression in multiple if / elif statements with comments.
Thanks!

Issue #475, opened 2019-08-01
Does onnx2tf support name scope?


JiayiFu:
Hi,
  I tried to use tensorboard to visualize the model converted from ONNX. But the model is too complex to understand. So if the model has some name scopes, the graph of the model may be more directly understanding. 
  Does onnx2tf support name scope in TensorFlow? like setting the 'domain' of ONNX op as the name scope of TensorFlow op? or some other ways?

fumihwh:
You can add something like `with tf.name_scope(node.name)` just after backend.py:L126 `for node in graph_def.node:`.

chinhuang007:
Since our handlers currently support domain=defs.ONNX_DOMAIN, which is '', the use of custom domains in ONNX ops for tf name scope will not work.

chinhuang007:
Please note node.name is optional in ONNX. So after `with tf.name_scope(node.name)` is added to backend.py, all nodes will be in the same tf name space if the exporter does not provide specific node names as "name space" indicators.

JiayiFu:
Thanks for your replies!

PR #524, opened 2019-11-12
Use the name of onnx node as the name scope of tensorflow.

JiayiFu:
As we discussed in #475. When we convert a model from ONNX to Tensorflow, we always use more than one ops to implement the converting. But these ops are named with something like "Equal_27", "Less_28", "add_46" which are impossible to know what layer or op them belong to. 
So I want to add the tf.name_scope to make the generated ops more clear.
ONNX doesn't support name scope now(onnx/onnx#2444) so I use the node's name as the name scope.


JiayiFu:
It seems like something wrong in the unit test env?

JiayiFu:
@chinhuang007 May you please review this PR?

chinhuang007:
@JiayiFu The idea of using "onnx node name" as the "tf name_scope" seems reasonable. @tjingrant What do you think? any other technical implications?

JiayiFu:
@tjingrant Some advice?

fumihwh:
Yes, as discussed in #475, add this sounds good.
But will it add some duplicate name? Such as 'xxx/xxx/conv/conv'?

JiayiFu:
> Yes, as discussed in #475, add this sounds good.
> But will it add some duplicate name? Such as 'xxx/xxx/conv/conv'?

Yes, if I name an ONNX Conv node as "conv_1". Then the names of generated TensorFlow ops will be something like 'conv_1/conv'. 
But I think the names like 'conv_1/conv' are better than the names we generate now: something like 'add_16', 'conv_28' and so on. At least we can know which layer the generated ops belong to. It is more clear and easy to find the errors if there is something wrong.

JiayiFu:
Close this. If I have any better solution I will create a new PR.

PR #528, opened 2019-11-22
Fix unknown shape from slice

chinhuang007:
As reported in issue 526, the slice would return output with
unknown shapes. The use of map_fn caused strided_slice unable
to do the proper shape inference.


chinhuang007:
This PR is part of the TF 2.0 PR, https://github.com/onnx/onnx-tensorflow/pull/531. Therefore close it.

PR #531, opened 2019-12-07
Support Tensorflow 2.0

chinhuang007:
Tensorflow 2.0 has a lot of API changes, in packaging, signature,
and also implementation.

This PR is to make sure the existing handlers can pass the node
tests. The use of compat.v1 APIs will be looked at in a separate
PR later.


chinhuang007:
@tjingrant @fumihwh Please review the changes to support Tensorflow 2.0. Once this PR is merged, the master branch will solely support 2.x and the tf-1.x branch is for 1.x going forward.

onnx_tf/handlers/backend/average_pool.py
```diff
@@ -25,9 +25,9 @@ def version_7(cls, node, **kwargs):
   @classmethod
   def version_10(cls, node, **kwargs):
     return cls.pool_v11(node, kwargs["tensor_dict"], "AVG",
-                    kwargs.get("strict", True))
+                        kwargs.get("strict", True))
```
fumihwh: just nit, yapf fomatter makes this?

test/backend/test_node.py
```diff
@@ -279,6 +279,7 @@ def test_constant_of_shape(self):
     np.testing.assert_almost_equal(output["Y"], np.zeros(x, dtype=np.int32))
 
   def test_conv(self):
+    import pdb; pdb.set_trace();
```
fumihwh: for DEBUG?

fumihwh:
Great work!
Just concern, so we should maintain two branch from now on until Google deprecate tf 1.x, :).............
I know the answer would be yes, but I can't find any way to avoid this now... sad

onnx_tf/handlers/backend/average_pool.py
chinhuang007: Yes, this is the result from yapf.

test/backend/test_node.py
chinhuang007: Good catch, just removed it.

chinhuang007:
Unfortunately we will need to maintain two branches for awhile :-(

chinhuang007:
@fumihwh @tjingrant could you please review and approve if everything looks okay? really would like to get this TF 2.0 support move forward. Thanks!

PR #537, opened 2020-01-07
add version 11 for Gather op

winston-zillow:


chinhuang007:
@winston-zillow Thanks for opening the PR. Getting an updated ONNX operator to work, , in this particular case, involves a little more, including handle negative and out-of-bound indices, provide test cases if needed, and update/regen the doc. You can find the general steps here (applicable to new and updated ops), https://github.com/onnx/onnx-tensorflow/blob/master/doc/IMPLEMENTING_NEW_OP.md

@winnietsang Since you are working on a similar PR, please help provide additional guidelines. Thanks!

winnietsang:
@winston-zillow Gather has two updates in Opset 11, (a) support negative indices and (b) if any index values are out of bounds then need to throw exception. I created a PR #538 to support similar features for GatherND. You may be able to reuse some of the code there. If you have any question please let me know.

winston-zillow:
> Gather has two updates in Opset 11, (a) support negative indices and (b) if any index values are out of bounds then need to throw exception

@winnietsang I just try to see if we can get the basic version support working similar to [version 11 for the `slice` op](https://github.com/onnx/onnx-tensorflow/blob/master/onnx_tf/handlers/backend/slice.py#L125). Will need to spend time working on the complete implementation

onnx_tf/handlers/backend/gather.py
```diff
@@ -12,3 +12,19 @@ class Gather(BackendHandler):
...
+    x_rank = len(x.shape)
+    if axis < -x_rank or axis >= x_rank:
+      raise ValueError("axis %d is out of range. must be in [%d, %d]" % (axis, -x_rank, x_rank-1))
+    dimension = tf.cast(tf.shape(x)[axis], indices.dtype)
+    indices_shape = tf.shape(indices)
+    dim_broadcasted = tf.broadcast_to(dimension, indices_shape)
+    nonneg_indices = tf.cast(tf.greater_equal(indices, 0), dtype=indices.dtype)
+    negative_indices = tf.cast(tf.less(indices, 0), dtype=indices.dtype)
+    final_indices = dim_broadcasted * negative_indices + negative_indices * indices + nonneg_indices * indices
+    return [tf.gather(x, final_indices, axis=axis, name=cls.tf_node_name(node.name))]
```
winston-zillow: @winnietsang here are the negative indices support. Note that other op transformation invoking `tf` functions do not set the op name; I set it here, though if that's against the convention I can remove it.

Also please specify what's needed for the error checking. Here I validate the `axis` attribute. For indices, the shape can be undefined and that run-time error will be thrown by tensorflow if the indices are out of range at run-time.

winnietsang: @winston-zillow please check indices out of bounds and throw exception before calling tf.gather. 
Tensorflow run in CPU will detect indices out of bounds and throw exception. But in GPU, Tensorflow will not throw exception, it will return a '0' for that corresponding output value. 
In Tensorflow documentation (https://www.tensorflow.org/api_docs/python/tf/gather) it state that this is the expected behavior. While in ONNX documentation (https://github.com/onnx/onnx/blob/master/docs/Operators.md#Gather) whenever there is indices out of bounds need to throw exception.

winston-zillow: @winnietsang do you see it is sufficient to raise out of bound error when the static shape is known. I wonder if there is a way to get TF to raise an error after checking the dynamic shapes.

winston-zillow: @winnietsang OK, I saw that you use `tf.compat.v1.assert_equal` in the `gather_nd` op. strange that the `assert_equal` became undocumented since TF 1.12, not even in the `t1.compat`. Will take use of that as well

winnietsang: @winston-zillow Yes, I used tf.compat.v1.assert_equal to get Tensorflow to throw exception for me in the gather_nd support, you can reuse all of those code here.

onnx_tf/handlers/backend/gather.py
```diff
@@ -12,3 +12,22 @@ class Gather(BackendHandler):
...
+    axis = node.attrs.get("axis", 0)
+    dimension = tf.cast(tf.shape(x)[axis], indices.dtype)
+    indices_shape = tf.shape(indices)
+    dim_broadcasted = tf.broadcast_to(dimension, indices_shape)
+    nonneg_indices = tf.cast(tf.greater_equal(indices, 0), dtype=indices.dtype)
+    negative_indices = tf.cast(tf.less(indices, 0), dtype=indices.dtype)
+    final_indices = dim_broadcasted * negative_indices + negative_indices * indices + nonneg_indices * indices
+    cond1 = tf.greater_equal(final_indices, 0)
+    cond2 = tf.less(final_indices, dimension)
+    indice_valiation = tf.reduce_all(tf.logical_and(cond1, cond2))
```
winston-zillow: @winnietsang for the `gather`, it is simpler than the `*_nd` ops and I try to avoid the `tf.while` invocation, so here I did the validation directly after computing the final indices.

winnietsang: @winston-zillow Please extend the Gather handler with GatherAndScatterMixin.
GatherAndScatterMixin is a class to provide common functions to handle tasks needed in Gather and Scatter related handlers 

For example  to check are there any indices are out of bounds can simply call 
result = cls.chk_idx_out_of_bounds_along_axis(data, axis, indices)

For process negative indices can call the following common function
indices = cls.process_neg_idx_along_axis(data, axis, indices)

winston-zillow: @winnietsang done. didn't notice the `*along_axis` versions.

onnx_tf/handlers/backend/gather.py
```diff
@@ -3,12 +3,24 @@
...
+  @classmethod
+  def version_11(cls, node, **kwargs):
+    x = kwargs["tensor_dict"][node.inputs[0]]
+    indices = kwargs["tensor_dict"][node.inputs[1]]
+    axis = node.attrs.get("axis", 0)
+    result = cls.chk_idx_out_of_bounds_along_axis(x, axis, indices)
+    msg = 'Gather indices are out of bounds, please double check the indices and retry.'
+    with tf.control_dependencies([tf.compat.v1.assert_equal(result, True, message=msg)]):
+      indices = cls.process_neg_idx_along_axis(x, axis, indices)
+      return [tf.gather(x, indices, axis=axis, name=cls.tf_node_name(node.name))]
```
winnietsang: @winston-zillow don't need to set name here. By default it will use the node.name as name

winston-zillow: @winnietsang are your sure that the `node.name` would be used automatically? How is the node name passed to the tensorflow op? it seems only the ops using `make_tensor_from_onnx_node` would inherit the node name.

winnietsang: @winston-zillow, you are right node.name will not be use because it call tf.gather directly here, lets change it to use cls.make_tensor_from_onnx_node, then the name will be set for you automatically.

winston-zillow: @winnietsang done. Note: this problem exists for other ops that do not use the `make_tensor_from_onnx_node`

onnx_tf/handlers/backend/gather.py
```diff
@@ -3,12 +3,25 @@
...
+  def version_11(cls, node, **kwargs):
+    x = kwargs["tensor_dict"][node.inputs[0]]
+    indices = kwargs["tensor_dict"][node.inputs[1]]
+    axis = node.attrs.get("axis", 0)
+    result = cls.chk_idx_out_of_bounds_along_axis(x, axis, indices)
+    msg = 'Gather indices are out of bounds, please double check the indices and retry.'
+    with tf.control_dependencies([tf.compat.v1.assert_equal(result, True, message=msg)]):
+      indices = cls.process_neg_idx_along_axis(x, axis, indices)
+      kwargs["tensor_dict"][node.inputs[1]] = indices
+      return [cls.make_tensor_from_onnx_node(node, **kwargs)]
```
winnietsang: @winston-zillow all the updated inputs and attrs need to send to cls.make_tensor_from_onnx_node.
You may use the Split https://github.com/onnx/onnx-tensorflow/blob/master/onnx_tf/handlers/backend/split.py handler as your reference here.

winnietsang: @winston-zillow if axis is not set then need to send the default value of axis to cls.make_tensor_from_onnx_node.
You may use the Split https://github.com/onnx/onnx-tensorflow/blob/master/onnx_tf/handlers/backend/split.py handler as a reference here.

onnx_tf/handlers/backend/gather.py
```diff
@@ -1,14 +1,30 @@
...
+  def version_11(cls, node, **kwargs):
+    x = kwargs["tensor_dict"][node.inputs[0]]
+    indices = kwargs["tensor_dict"][node.inputs[1]]
+    attrs = copy.deepcopy(node.attrs)
+    axis = attrs.get("axis", 0)
+    result = cls.chk_idx_out_of_bounds_along_axis(x, axis, indices)
+    msg = 'Gather indices are out of bounds, please double check the indices and retry.'
+    with tf.control_dependencies([tf.compat.v1.assert_equal(result, True, message=msg)]):
+      indices = cls.process_neg_idx_along_axis(x, axis, indices)
+      kwargs["tensor_dict"][node.inputs[1]] = indices
```
chinhuang007: @winston-zillow Please don't change kwargs. Send in specific inputs when necessary.

chinhuang007:
@winston-zillow We always run test_onnx_back.py to ensure the PR works everywhere. It will catch the error of updating kwargs. You can see the same same problem in CI report as well.

winnietsang:
LGTM!

PR #539, opened 2020-01-10
Add opset 11 support for Conv

chinhuang007:
Add opset 11 support for Conv and update test code for
conv to run on GPU and CPU machines.
Also regenerated the status file for TF 2.x


chinhuang007:
The Conv and ConvTranspose for opset 11 added only the documentation of "dilations and strides defaults to 1". Our code already made the same assumption. @tjingrant @fumihwh  please review and make sure I did not miss anything.

dipsatch:
I am getting `NotImplementedError: Gemm version 11 is not implemented.`, when I run: `onnx-tf convert -i model.onnx -o model.onnx_to_tf_output.pb`. Can you please help?

PR #541, opened 2020-01-13
Update Conv and ConvTranspose for opset 11

chinhuang007:
Update Conv and ConvTranspose handlers for opset 11 and
test code for Conv. Similar changes can be refereced at
PR https://github.com/onnx/onnx-tensorflow/pull/539


chinhuang007:
@tjingrant @fumihwh Same fix as the recent merged PR for tf 2.0. Please review. Thanks!

PR #544, opened 2020-01-23
preserve names of initial tensor constants

winston-zillow:
When a ONNX graph is converted, the initialization constant tensor names are not preserved in the target tensorflow model. Preserving constant names should improve quality of the converted model.

https://github.com/onnx/onnx-tensorflow/issues/545


winston-zillow:
@winnietsang @chinhuang007 please review

Issue #516, opened 2019-10-18
NotImplementedError: Split version 11 is not implemented.


ningningG:
Platform (like ubuntu 16.04/win10):
win10
Python version:
3.7.4
Tensorflow version:
Tensorflow 1.14.0 with GPU

There is  an NotImplementedError when converte  onnx_node to tensorflow_op:

```
Traceback (most recent call last):
  File "D:/mx2tf_CT/mx2tf/onnx2tf1.py", line 6, in <module>
    tf_rep = prepare(onnx_model)  # prepare tf representation
  File "C:\Users\Administrator\AppData\Local\Continuum\anaconda3\envs\model_convertor\lib\site-packages\onnx_tf\backend.py", line 55, in prepare
    return cls.onnx_model_to_tensorflow_rep(model, strict)
  File "C:\Users\Administrator\AppData\Local\Continuum\anaconda3\envs\model_convertor\lib\site-packages\onnx_tf\backend.py", line 75, in onnx_model_to_tensorflow_rep
    return cls._onnx_graph_to_tensorflow_rep(model.graph, opset_import, strict)
  File "C:\Users\Administrator\AppData\Local\Continuum\anaconda3\envs\model_convertor\lib\site-packages\onnx_tf\backend.py", line 129, in _onnx_graph_to_tensorflow_rep
    onnx_node, tensor_dict, handlers, opset=opset, strict=strict)
  File "C:\Users\Administrator\AppData\Local\Continuum\anaconda3\envs\model_convertor\lib\site-packages\onnx_tf\backend.py", line 226, in _onnx_node_to_tensorflow_op
    return handler.handle(node, tensor_dict=tensor_dict, strict=strict)
  File "C:\Users\Administrator\AppData\Local\Continuum\anaconda3\envs\model_convertor\lib\site-packages\onnx_tf\handlers\handler.py", line 63, in handle
    exception.OP_UNIMPLEMENTED_EXCEPT(node.op_type, cls.SINCE_VERSION)
  File "C:\Users\Administrator\AppData\Local\Continuum\anaconda3\envs\model_convertor\lib\site-packages\onnx_tf\common\exception.py", line 34, in __call__
    super(OpUnimplementedException, self).__call__(op, version, domain)
  File "C:\Users\Administrator\AppData\Local\Continuum\anaconda3\envs\model_convertor\lib\site-packages\onnx_tf\common\exception.py", line 13, in __call__
    raise self._func(self.get_message(*args, **kwargs))
[...]

PR #540, opened 2020-01-13
Add opset 11 support for Split

chinhuang007:
Add opset 11 support for Split in checking the axis to be
inbound and providing a test case.

PR #549, opened 2020-01-25
Update Split for opset 11

chinhuang007:
Update Split handler and test code for opset 11.
Similar changes can be refereced at
https://github.com/onnx/onnx-tensorflow/pull/540


pluradj:
Fixes #516

PR #551, opened 2020-01-29
tf-1.x: preserve names of initial tensor constants

winston-zillow:
Apply the change in https://github.com/onnx/onnx-tensorflow/pull/544 to tf-1.x branch


winston-zillow:
@chinhuang007 please approve and merge, so this can be closed

Issue #542, opened 2020-01-21
NotImplementedError: Gemm version 11 is not implemented.


futuyao:
**Describe the bug**

I'm trying to convert an onnx model into tensorflow, and getting "`NotImplementedError: Gemm version 11 is not implemented`" error.

**To Reproduce**

Running conversion from command line with: `onnx-tf convert -i GoogLeNetPlaces.onnx -o GoogLeNetPlaces.pb`

**ONNX model file**

Here's the model I'm trying to convert: 
https://drive.google.com/file/d/1CfZTL7vl9HaeApIYHYbAWyIetPAPsf2c/view

**Python, ONNX, ONNX-TF, Tensorflow version**

This section can be obtained by running `get_version.py` from util folder.
 - Python version: 3.7.6
 - ONNX version: 1.6.0
 - ONNX-TF version: from source (commit ddfea34f1fa528c4d56ed1a3d070af019fac58db)
 - Tensorflow version: 2.1.0

**Additional context**

Ehtasha:
I'm also facing this issue.
Did you find any solution ?

futuyao:
@Ehtasha Not that I know of at the moment.

pluradj:
#554 was merged recently on `master`

pluradj:
See also #558 for `tf-1.x`

futuyao:
Thanks!

Hari-Guhan-s:
i am getting the same issue in rel1.3

chinhuang007:
@Hari-Guhan-s The patch is in the master and tf-1.x branches. It is not officially available yet via pip install. To pick up the fix, please do a source install, https://github.com/onnx/onnx-tensorflow#installation

Hari-Guhan-s:
Thanks

Hari-Guhan-s:
> Thanks @chinhuang007  but as I am using tensorflow 1.14 I cannot move to master branch or tf1.x branch and I did a source installation from rel1.3 

Any update on this issue

PR #554, opened 2020-02-04
Add opset 11 support for Gemm

pluradj:
Input tensor C is now optional. If not specified, the computation is done
as if C is a scalar 0.

---

A new test case `test_gemm_default_no_bias` was previously added to `onnx` to handle the new functionality. I verified that `onnx-tensorflow` passes all Gemm backend tests with `test_onnx_backend.py` and `test_node.py`.

I ran `gen_status.py -v master` with onnx master (opset 12) installed. I noticed that the latest commit on onnx-tensorflow master for [Gather](https://github.com/onnx/onnx-tensorflow/commit/abf3e29d7232fa1da63eb2cb0a0bf300ff36fb66) op was likely run against onnx v1.6.0 (opset 11).


pluradj:
Fixes #542

PR #558, opened 2020-02-06
Add opset 11 support for Gemm

pluradj:
Input tensor C is now optional. If not specified, the computation is done
as if C is a scalar 0.

---

Same fix as #554 targeted for TensorFlow 1.x branch.


pluradj:
Fixes #542 for `tf-1.x` branch

PR #564, opened 2020-02-12
Add opset 11 support for Det

pluradj:

---

I verified that `onnx-tensorflow` passes all `Det` backend tests with `test_onnx_backend.py`


onnx_tf/handlers/backend/det.py
```diff
@@ -0,0 +1,18 @@
...
+@onnx_op("Det")
+@tf_func(tf.linalg.det)
+class Det(BackendHandler):
+
+  @classmethod
+  def version_11(cls, node, **kwargs):
+    tensor_dict = kwargs["tensor_dict"]
+    x = tensor_dict[node.inputs[0]]
+    return [
+        cls.make_tensor_from_onnx_node(node, inputs=[x])
```
chinhuang007: Just a nit. Could code be simplified as cls.make_tensor_from_onnx_node(node, **kwargs)?

pluradj: Thanks, updated.

PR #565, opened 2020-02-12
Dynamic shape, v12 support and removed legacy code for MaxPool and AveragePool

sdmonov:
Also added v12 support for MaxPool (support for int8 and uint8 dtypes)

Removed old code that was used for v8 or less of the ops.
There are two things need to be considered between the new and old code:

1) The old implementation would execute the tf ops with channel first if
   there is GPU on the current machine. This will be considered as a separate
   PR part of bigger effort to generalize the behaviour to support channel
   first, channel last for all ops

2) The old implementation will check if the explicit pads provided match
   "SAME" padding in TF. If it matches will not use tf.pad but instead will
   use padding="SAME" when calling the tf operators. This code uses get_shape()
   to do the calculations. The current PR maintains same approach only if the shape
   is fully defined. In case of unknown shape it will use explicit padding always.

Both of the above features are optimizations for performance and do not
reflect the results.


chinhuang007:
@sdmonov @winnietsang Please make sure this is added to the tf-1.x branch as well. Thanks.

PR #567, opened 2020-02-13
support `floormod` in older tf 1.x versions

winston-zillow:
For tf version older than 1.15, the `floormod` is not in the `tf.math` module.


chinhuang007:
@winston-zillow please resolve the conflicts.

PR #568, opened 2020-02-14
Add opset 11 support for DynamicQuantizeLinear

pluradj:

---

I verified that `onnx-tensorflow` passes all `DynamicQuantizeLinear` backend tests (6) with `test_onnx_backend.py`

I ran the `yapf` command against `onnx_tf/opset_version.py` and `test/backend/test_node.py` so you'll see some extra reformatting included in those file in addition to the expected changes for this operator.


pluradj:
I noticed with the latest versions of `onnx` master branch, some test cases were being skipped:

- test_dynamicquantizelinear_max_adjusted_cpu
- test_dynamicquantizelinear_max_adjusted_expanded_cpu
- test_dynamicquantizelinear_min_adjusted_cpu
- test_dynamicquantizelinear_min_adjusted_expanded_cpu
- test_clip_default_max_cpu
- test_clip_default_min_cpu
- test_operator_max_cpu
- test_operator_maxpool_cpu
- test_operator_min_cpu

There were some Opset 12 changes on the operators `Max` and `Min`. It looks like the exclusion filter in `test_onnx_backend.py` was being overly aggressive on `_max_` and `_min_`. The fix I put in place will match specifically on `test_opname` rather than `[a-z,_]*_opname`.

chinhuang007:
@winnietsang This looks good to me. Please review the changes in test_onnx_backend.py and make sure the filtering logic properly skips the unsupported tests.

winnietsang:
> @winnietsang This looks good to me. Please review the changes in test_onnx_backend.py and make sure the filtering logic properly skips the unsupported tests.

@chinhuang007 there is no official naming convention on testcase name in ONNX. Most of the test are using name like this "test_opname_" then some other description text after the opname but some of the test are like this begin with "test_" then some description text and then follow by "opname" then more description_text after the opname. Therefore there is always chances that it may skip some test unintentionally or some test suppose to be skip but it didn't skip. This is just a best effort we can do here and the list of ops that need to skip always change because we continue to add support for new/updated ops into our convertor and at the same time ONNX is also continue to add new/updated ops into ONNX too. I agree with the changes in test_onnx_backend.py here. I think this change will make some test suppose to skip but will not skip now. I think that's OK. But I think in the future we need to verify that the onnx_backend test for the op that your PR is updating is not skip. If it is skipped then need to change this logic again.

chinhuang007:
I agree we should skip tests on the conservative side rather than overly aggressive. If we see the tests that should be skipped showing as errors, we will update the current or write new filters.

PR #572, opened 2020-02-17
Add opset 11 support for Clip (tf-1.x)

pluradj:
min and max moved from required attributes to optional inputs

---

Adding `Clip` support on `tf-1.x` branch, same changes as #556.


chinhuang007:
Looks good. Please regenerate the status report.

PR #575, opened 2020-02-18
Add opset 11 support for CumSum

pluradj:

---

I verified that `onnx-tensorflow` passes all `CumSum` backend tests (7) with `test_onnx_backend.py`. Also note, I've opened related PR https://github.com/onnx/onnx/pull/2611 against `onnx` and verified that this PR works with optional `axis`.


onnx_tf/handlers/backend/cumsum.py
```diff
@@ -0,0 +1,45 @@
...
+  def version_11(cls, node, **kwargs):
+    tensor_dict = kwargs["tensor_dict"]
+    x = tensor_dict[node.inputs[0]]
+    inputs = [x]
+
+    if len(node.inputs) > 1:
+      # optional 0-D tensor, range [-rank(x), rank(x)-1]
+      axis = tensor_dict["axis"]
+      inputs.append(tf.reduce_min(axis))
+
```
chinhuang007, line 39: tf.reduce_min doesn't seem necessary here. The input axis is already a 0-D tensor, meaning the number of elements always 1.

chinhuang007: @pluradj can you check whether tf.reduce_min is needed and provide comments? Thanks!

pluradj: You're right @chinhuang007 -- the input is declared in the spec to be a 0-D tensor. I removed the call to `tf.reduce_min`.

The remaining problem is that the test cases in the `ONNX` repository were passing a 1-D tensor. I've pushed a fix to https://github.com/onnx/onnx/pull/2611 to fix that issue. Unfortunately, the CumSum `test_onnx_backend.py` tests won't pass until that PR is merged. I've tested with both PRs locally, and the backend test will pass.

Issue #584, opened 2020-03-10
Support negative inputs for 'Pad' op


fthielke:
**Describe the bug**

If I have a Pad op with negative paddings (in this case -16 in all the spatial dimensions of a 3D volume) in my model, `onnx_tf.backend.prepare` raises the following error:
```
ValueError: Paddings must be non-negative for 'Pad' (op: 'Pad') with input shapes: [1,32,40,40,40], [5,2] and with computed input tensors: input[1] = <[0 0][0 0][-16 -16][-16 -16][-16 -16]>.
```

However, this is not in line with the ONNX spec which supports negative paddings (ie. croppings) since version 2: https://github.com/onnx/onnx/blob/master/docs/Changelog.md#Pad-2

**Python, ONNX, ONNX-TF, Tensorflow version**

 - Python version: 3.7.6 (default, Jan  8 2020, 20:23:39) [MSC v.1916 64 bit (AMD64)]
 - ONNX version: 1.5.0
 - ONNX-TF version: 1.5.0
 - Tensorflow version: 2.1.0

chinhuang007:
Thanks for opening the issue. It is definitely a bug. We will look into a solution. Please note the converted graph will be much bigger since Tensorflow pad operator does not support negative pads.

PR #588, opened 2020-03-27
Add negative paddings support for pad

chinhuang007:
This PR is to add negative paddings support for pad operator
according to ONNX spec, negative pads mean to remove elements.

It also should address the issue
https://github.com/onnx/onnx-tensorflow/issues/584

The current assumption is that paddings can either positive or
negative, not mix of them, since ONNX spec does not define
such scenario and behavior.


tjingrant:
Hi Chin, how does this approach compare with, say using `tf.slice`? It seems like you have to materialize the coordinate vector for each pixel that is not removed by negative padding, and using `tf.gather` also carries performance penalties, given that it's fundamentally a sparse operation.

chinhuang007:
Thanks for the review, Tian. The logic is similar to tf.slice. However getting the proper begins and sizes out of the 1-d tensor pads and input shape requires calculation (including gather I believe) as well. Would you say using tf.slice will definitely improve performance? I could take a look at the alternative.

chinhuang007:
@tjingrant I changed the code to use tf.slice. It is much simpler. Thanks very much for your suggestion. Please review.

onnx_tf/handlers/backend/pad.py
```diff
@@ -17,40 +16,63 @@ def _common(cls, node, **kwargs):
...
+        return i + 1, x
+
+      # tf requires int32 paddings
+      paddings = tf.cast(tf.transpose(tf.reshape(paddings, [2, num_dim])),
+                         dtype=tf.int32)
+
+      if mode.lower() == "edge":
+        paddings = tf.reshape(paddings, [-1])
+        max_i = tf.reduce_max(paddings)
+        _, x = tf.while_loop(
```
tjingrant, line 53: Hi Chin, I either forgot or never understood what this part is doing, could you give me some hints about what's going on with the while loop? Some comments may help with future reviews, too.

chinhuang007: The while loop code is not changed in this PR, just added with indentation. It does look a bit strange. I will need to investigate what it is doing and whether it works as expected.

chinhuang007: Okay, now I understand the while loop is to replace the python implementation of edge mode, np.pad(x, pads, mode="edge"). Basically it goes through the tf.pad 'SYMMETRIC' mode to pad one value at a time for both sides and all dimensions. I added some comments and hope it helps.

Issue #600, opened 2020-04-16
TypeError: Value passed to parameter 'x' has DataType bool not in list of allowed values: bfloat16, float16, float32, float64


heliqi:
**Describe the bug**

While trying to convert the model from onnx to tensorflow i encountered the following issue

TypeError: Value passed to parameter 'x' has DataType bool not in list of allowed values: bfloat16, float16, float32, float64 ...etc

**Python, ONNX, ONNX-TF, Tensorflow version**

This section can be obtained by running `get_version.py` from util folder.
 - Python version: 3.6.8
 - ONNX version:  1.6.0
 - ONNX-TF version:1.5.0
 - Tensorflow version: 1.15.0
 
Note:
Error info  show: this error occurred in the tensorflow_core/python/framework/op_def_library.py code. 
error op is  "mul",  describe 'x' must be 'bfloat16, float16, float32, float64 ...etc', not be  bool.

How can I solve this problem？

heliqi:
Through print info, I find "Expand" has a problem. When x.dtyoe is tf.bool,  'x * ones' is an error.  Because the input of tf.Mul must not be  tf.bool type.
```
@onnx_op("Expand")
class Expand(BackendHandler):

  @classmethod
  def version_8(cls, node, **kwargs):
    tensor_dict = kwargs["tensor_dict"]
    x, shape = tensor_dict[node.inputs[0]], tensor_dict[node.inputs[1]]
    ones = tf.ones(shape, dtype=x.dtype)
    return [x * ones]
```
https://github.com/onnx/onnx-tensorflow/blob/c4fc047e7ec71daa6aa8f71e9cc2ee9e5a3768b6/onnx_tf/handlers/backend/expand.py#L15

chinhuang007:
Good catch! A patch coming soon.

GotG:
hello, i pip installed onnx-tf and i still get this error:
'TypeError: Value passed to parameter 'input' has DataType bool not in list of allowed values: float32, float64, int32, uint8, int16, int8, complex64, int64, qint8, quint8, qint32, bfloat16, uint16, complex128, float16, uint32, uint64'

PR #602, opened 2020-04-16
Fix an expand issue with bool

chinhuang007:
If the input data type is bool, the expand operator would
fail to convert and throw an exception, as described in
https://github.com/onnx/onnx-tensorflow/issues/600

This PR converts input data to tf.int32 for matmul when
dtype is bool.

### DIFF
--- a/onnx_tf/handlers/backend/expand.py
+++ b/onnx_tf/handlers/backend/expand.py
@@ -11,5 +11,12 @@ class Expand(BackendHandler):
   def version_8(cls, node, **kwargs):
     tensor_dict = kwargs["tensor_dict"]
     x, shape = tensor_dict[node.inputs[0]], tensor_dict[node.inputs[1]]
-    ones = tf.ones(shape, dtype=x.dtype)
-    return [x * ones]
+
+    # tf.math.multiply does not support bool therefore use int8
+    if x.dtype is tf.bool:
+      ones = tf.ones(shape, dtype=tf.int8)
+      r = tf.cast(x, tf.int8) * ones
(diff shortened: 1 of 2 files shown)


chinhuang007:
@tjingrant @fumihwh Could you please review this simple fix to resolve an open user issue?

tjingrant:
Hi Chin, how does this approach compare with, say using `tf.slice`? It seems like you have to materialize the coordinate vector for each pixel that is not removed by negative padding, and using `tf.gather` also carries performance penalties, given that it's fundamentally a sparse operation.

onnx_tf/handlers/backend/expand.py
```diff
@@ -11,5 +11,12 @@ class Expand(BackendHandler):
   def version_8(cls, node, **kwargs):
     tensor_dict = kwargs["tensor_dict"]
     x, shape = tensor_dict[node.inputs[0]], tensor_dict[node.inputs[1]]
-    ones = tf.ones(shape, dtype=x.dtype)
-    return [x * ones]
+
+    # tf.matmul does not support bool therefore use int32
```
tjingrant, line 15: Why is tf.matmul relevant here?

chinhuang007: Yeah, wrong comment, it is tf.math.multiply, not matmul.

tjingrant:
Also just minor suggestion, int32 expands the memory footprint by at least 4x. Using int8 may be more appropriate if it's possible & easy to do so.

chinhuang007:
Makes sense to use smaller int8. Thanks for the suggestion.

Issue #595, opened 2020-04-08
onnx_tf creates a Dilation2d layer which is not supported by TensorFlow Lite


devidlatkin:
I use a docker image from here https://github.com/onnx/onnx-tensorflow/blob/master/README.md to transform my image from PyTorch to ONNX first, then to Frozen Graph and to tflite after. 
After transforming PyTorch to ONNX I have the next model summary acquired from print(onnx.helper.printable_graph(model_onnx.graph)) :

graph torch-jit-export (
  %test_input[FLOAT, 1x1x240x320]
) initializers (
  %conv1a.bias[FLOAT, 64]
  %conv1a.weight[FLOAT, 64x1x3x3]
  %conv1b.bias[FLOAT, 64]
  %conv1b.weight[FLOAT, 64x64x3x3]
  %conv2a.bias[FLOAT, 64]
  %conv2a.weight[FLOAT, 64x64x3x3]
  %conv2b.bias[FLOAT, 64]
  %conv2b.weight[FLOAT, 64x64x3x3]
  %conv3a.bias[FLOAT, 128]
  %conv3a.weight[FLOAT, 128x64x3x3]
  %conv3b.bias[FLOAT, 128]
  %conv3b.weight[FLOAT, 128x128x3x3]
  %conv4a.bias[FLOAT, 128]
  %conv4a.weight[FLOAT, 128x128x3x3]
  %conv4b.bias[FLOAT, 128]
  %conv4b.weight[FLOAT, 128x128x3x3]
  %convDa.bias[FLOAT, 256]
  %convDa.weight[FLOAT, 256x128x3x3]
  %convDb.bias[FLOAT, 256]
  %convDb.weight[FLOAT, 256x256x1x1]
  %convPa.bias[FLOAT, 256]
  %convPa.weight[FLOAT, 256x128x3x3]
  %convPb.bias[FLOAT, 65]
  %convPb.weight[FLOAT, 65x256x1x1]
) {
  %25 = Conv[dilations = [1, 1], group = 1, kernel_shape = [3, 3], pads = [1, 1, 1, 1], strides = [1, 1]](%input, %conv1a.weight, %conv1a.bias)
  %26 = Relu(%25)
  %27 = Conv[dilations = [1, 1], group = 1, kernel_shape = [3, 3], pads = [1, 1, 1, 1], strides = [1, 1]](%26, %conv1b.weight, %conv1b.bias)
  %28 = Relu(%27)
[...]

PR #613, opened 2020-05-05
Do not use tf.nn.dilation2d with dalations==1. use tf.nn.pool instead.

sdmonov:
When not needed, do not use dilation2d operation (in case of dilations==1) and use more general tf.nn.pool operation instead.

Fixes: #595


sdmonov:
@Wheest thanks for reviewing. I reverted the formatting changes. You are right, this should be done separately.

Issue #618, opened 2020-05-17
tensorflow.python.framework.errors_impl.InvalidArgumentError: Dimensions must be equal


jaggernaut007:
**Describe the bug**

Getting this error while converting onnx to tensorflow.
this model works fine on pytorch. Seems like there is an issue with the layer 134.

**To Reproduce**

```
import onnx

from onnx_tf.backend import prepare

onnx_model = onnx.load("../margipose/pretrained/margipose2.onnx")  # load onnx model
tf_rep = prepare(onnx_model)  # prepare tf representation
tf_rep.export_graph("output_path")  # export the model
```

If a model exported from PyTorch and Caffe2 is having trouble in ONNX-TF, use the next section to attach the model.

**ONNX model file**

Link to the Py torch model 
https://cloudstor.aarnet.edu.au/plus/s/fg5CCss8o9PdURs

Link to onnx model
https://drive.google.com/file/d/1_o_6Gg6arMN4CqKEvaaNPQ-jI3fbcbIV/view?usp=sharing

**Python, ONNX, ONNX-TF, Tensorflow version**

This section can be obtained by running `get_version.py` from util folder.
 - Python version: 3.8
 - ONNX version: 1.7
 - ONNX-TF version: 1.5
 - Tensorflow version: 2.2

**Additional context**

Traceback (most recent call last):
  File "example/onnx_to_tf.py", line 6, in <module>
    tf_rep = prepare(onnx_model)  # prepare tf representation
  File "/home/shreyas/onnx/onnx-tensorflow/onnx_tf/backend.py", line 66, in prepare
    return cls.onnx_model_to_tensorflow_rep(model, strict)
  File "/home/shreyas/onnx/onnx-tensorflow/onnx_tf/backend.py", line 86, in onnx_model_to_tensorflow_rep
    return cls._onnx_graph_to_tensorflow_rep(model.graph, opset_import, strict)
  File "/home/shreyas/onnx/onnx-tensorflow/onnx_tf/backend.py", line 143, in _onnx_graph_to_tensorflow_rep
    output_ops = cls._onnx_node_to_tensorflow_op(onnx_node,
  File "/home/shreyas/onnx/onnx-tensorflow/onnx_tf/backend.py", line 249, in _onnx_node_to_tensorflow_op
    return handler.handle(node, tensor_dict=tensor_dict, strict=strict)
[...]

PR #622, opened 2020-05-21
Fix conv transpose shape calculation

chinhuang007:
The output shape calculation was not correctly following the
ONNX spec, https://github.com/onnx/onnx/blob/master/docs/Operators.md#convtranspose

The PR should fix issue https://github.com/onnx/onnx-tensorflow/issues/618

The other changes are the result of running yaph formatter.


jaggernaut007:
This works great|! Thanks for the fix @chinhuang007 . Appreciate it!

chinhuang007:
@tjingrant @fumihwh could you please review? Thanks.

PR #630, opened 2020-05-27
Add SplitToSequence Support for Opset 11

chinhuang007:
Add SplitToSequence Support for Opset 11


chinhuang007:
@pluradj Good catch. I just added the support for keepdims and a unit test for it. Please review again. Thanks!

Issue #639, opened 2020-06-04
Can't Prepare tensorflow model from onnx


akshayklr057:
**Describe the bug**
I was trying to convert a huggingface model to tensorflow. Getting error in "prepare" method of onnx_tf.backend. Getting a key error 'axis', attached below is the entire traceback of error and my model.

I converted a huggingface model trained & saved using pytorch to onnx version, so that I can convert it to tensorflow. 

**To Reproduce**

Please give us instructions to reproduce your problem.

A __self-contained__ piece of code that can demonstrate the problem is required.
`import onnx

from onnx_tf.backend import prepare

onnx_model = onnx.load("./version3.onnx")  # load onnx model
tf_rep = prepare(onnx_model)  # prepare tf representation
tf_rep.export_graph("version3.pb")`

Please do not expect us to have PyTorch, Caffe2 installed.

If a model exported from PyTorch and Caffe2 is having trouble in ONNX-TF, use the next section to attach the model.

**ONNX model file**
My model is hosted here.
https://drive.google.com/file/d/1oYB7cpz9qwtlR0nfIjW7HnIKpzQvlN94/view?usp=sharing

If applicable, attach the onnx model file in question using Gist, DropBox or Google Drive.

**Python, ONNX, ONNX-TF, Tensorflow version**
Python - 3.6
tensorflow - 2.2

This section can be obtained by running `get_version.py` from util folder.
 - Python version: 
 - ONNX version: 
 - ONNX-TF version:
 - Tensorflow version: 

**Additional context**

`2020-06-04 09:23:25,227 - onnx-tf - INFO - Unknown op ConstantFill in domain `ai.onnx`.
2020-06-04 09:23:25,228 - onnx-tf - INFO - Unknown op ImageScaler in domain `ai.onnx`.
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-37-0b06f55f6108> in <module>()
      6 
      7 onnx_model = onnx.load("./version3.onnx")  # load onnx model
----> 8 tf_rep = prepare(onnx_model)  # prepare tf representation
      9 tf_rep.export_graph("version3.pb")

5 frames
/content/onnx-tensorflow/onnx_tf/backend.py in prepare(cls, model, device, strict, logging_level, **kwargs)
     64     common.logger.handlers[0].setLevel(logging_level)
     65 
---> 66     return cls.onnx_model_to_tensorflow_rep(model, strict)
     67 
     68   @classmethod

[...]

PR #640, opened 2020-06-05
Fix cumsum input name issue

chinhuang007:
This PR is to fix the cumsum input name issue, as seen in
https://github.com/onnx/onnx-tensorflow/issues/639

The axis input can be provided with different actual names.
So we need to use input position instead of specific name.


winnietsang:
LGTM.

PR #642, opened 2020-06-05
Added Opset 13 support for ArgMax and ArgMin

shahirdaya:
The change in Opset 13 was a Type Constraint to support tensor(bfloat16). I have added test for this.

### DIFF
--- a/onnx_tf/handlers/backend/cast.py
+++ b/onnx_tf/handlers/backend/cast.py
@@ -6,12 +6,12 @@ from onnx_tf.handlers.handler import tf_func
 from onnx_tf.handlers.handler import partial_support
 from onnx_tf.handlers.handler import ps_description
 
-
 @onnx_op("Cast")
 @tf_func(tf.cast)
 @partial_support(True)
-@ps_description("Cast string to float32/float64/int32/int64 " +
-                "are not supported in Tensorflow.")
+@ps_description("Cast string to data types other than " +
+                "float32/float64/int32/int64 is not supported in Tensorflow")
+
 class Cast(BackendHandler):
 
   @classmethod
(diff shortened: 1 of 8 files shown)


test/backend/test_node.py
```diff
@@ -113,6 +114,22 @@ def test_arg_max(self):
...
+      if not legacy_opset_pre_ver(13):
+        node_def = helper.make_node("ArgMax", ["data"], ["reduced"],
+                                    axis=axis,
+                                    keepdims=0)
+        
+        np_fp32 = self._get_rnd_float32(shape=[10, 10])
+        little_endisan = sys.byteorder == 'little'
+        np_uint16_view = np_fp32.view(dtype=np.uint16)
+        np_bfp16 = np_uint16_view[1::2] if little_endisan else np_uint16_view[0::2]
+        data = np_bfp16
```
chinhuang007, line 127: The data seems in np.uint16. Can you please elaborate how this is going to test and verify bfloat16 handling in the handler?

shahirdaya: @chinhuang007 bfloat16 is very similar to float32 except that it has less precision. So the approach which I leveraged from Cast here: https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/cast.py (see lines 41 to 46) basically starts with a float32 and construct a view of the memory but as uint16. Once we do that, we need take care of the byte order. Basically manually creating a bfloat16. Hope that helps. Thanks

shahirdaya: @chinhuang007 thanks on the discussion on support for bfloat16. I am reworking this. Thanks again.

chinhuang007:
Looks great! Wonder if you can update the cast handler's partial support statement? It is not changed in the PR but seems incorrect. Since cast.py is part of this PR, please help correct the statement. Thanks!

shahirdaya:
@chinhuang007 you're right. The partial support statement for cast is no longer true.  I have removed it and amended my commit.  Thanks for catching that.

onnx_tf/handlers/backend/cast.py
```diff
@@ -3,31 +3,18 @@
...
 from onnx_tf.handlers.handler import tf_func
-from onnx_tf.handlers.handler import partial_support
-from onnx_tf.handlers.handler import ps_description
-
 
 @onnx_op("Cast")
 @tf_func(tf.cast)
-@partial_support(True)
-@ps_description("Cast string to float32/float64/int32/int64 " +
-                "are not supported in Tensorflow.")
```
chinhuang007, line 14: Sorry for the delayed review. I believe we still have limited support because cast from a string to another data type depends on tf.strings.to_number, which supports four output types, tf.float32, tf.float64, tf.int32, tf.int64. The currently description states the opposite. Maybe change the wording to something like "Cast string to data types other than float32/float64/int32/int64 is not supported in Tensorflow"?

shahirdaya: You're right @chinhuang007. I should have caught that.  We are dependent on tf.string.to_number for now.  I have made the update. Thanks

PR #643, opened 2020-06-05
Fix initializer node name with leading underscore character

chinhuang007:
Node name cannot start with an underscore character in Tensorflow.
This PR prepends a unique string to make it unique and valid.


winnietsang:
LGTM

PR #654, opened 2020-06-11
Fix map_fn with string issue on gpu

chinhuang007:
The tfidf vectorizer handler has a map_fn with string constant
that doesn't work on gpu only. It is a known issue in Tensorflow.
This PR is to provide a temporary fix by placing the map_fn
specifically on cpu. Once the TF bug is fixed, we can remove
the fix.

### DIFF
--- a/onnx_tf/handlers/backend/tfidf_vectorizer.py
+++ b/onnx_tf/handlers/backend/tfidf_vectorizer.py
@@ -42,7 +42,18 @@ class TfIdfVectorizer(BackendHandler):
     # Loop through the ngram targets in the pool
     tensor_list = []
     for i in range(len(pool)):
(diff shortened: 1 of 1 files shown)


onnx_tf/handlers/backend/tfidf_vectorizer.py
```diff
@@ -42,7 +42,13 @@ def _calc_ngram_skip(cls, x, pool, n, skip=0):
     # Loop through the ngram targets in the pool
     tensor_list = []
     for i in range(len(pool)):
-      ngram_count = tf.map_fn(lambda in_x: tf.where(tf.reduce_all(tf.equal(in_x, tf.constant(pool[i], dtype=new_x.dtype))), tf.constant([1]), tf.constant([0])), new_x, dtype=tf.int32)
+      with tf.device("/cpu:0"):
+        ngram_count = tf.map_fn(lambda in_x: tf.where(
```
winnietsang, line 46: May be add some comment to explain why need to run this map_fn on cpu device

chinhuang007: Thanks for the review. Added some comments. Please review again.

chinhuang007:
Good suggestion. The issue link is added to the comment.

PR #661, opened 2020-06-12
Add opset 11 support for GatherElements

pluradj:

---

* Tests added to `test_node.py`
* Verified `GatherElements` backend tests pass (3) with `test/backend/test_onnx_backend.py`
* Ran `yapf` against `onnx_tf/opset_version.py` so there is some additional reformatting in that file

### DIFF
--- a/onnx_tf/handlers/backend/gather_elements.py
+++ b/onnx_tf/handlers/backend/gather_elements.py
@@ -1,0 +1,58 @@
+import tensorflow as tf
+
+from onnx_tf.handlers.backend_handler import BackendHandler
+from onnx_tf.handlers.handler import onnx_op
+from .gather_and_scatter_mixin import GatherAndScatterMixin
+
+
+@onnx_op("GatherElements")
+class GatherElements(GatherAndScatterMixin, BackendHandler):
+
+  @classmethod
+  def version_11(cls, node, **kwargs):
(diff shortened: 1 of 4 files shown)


onnx_tf/handlers/backend/gather_elements.py
```diff
@@ -0,0 +1,56 @@
...
+          map(lambda x: tf.range(x, dtype=index_swaped.dtype),
+              index_swaped.shape.as_list())),
+                                         indexing='ij')
+      idx_tensors_per_axis[0] = index_swaped
+      dim_expanded_idx_tensors_per_axis = list(
+          map(lambda x: tf.expand_dims(x, axis=-1), idx_tensors_per_axis))
+      index_expanded = tf.concat(dim_expanded_idx_tensors_per_axis, axis=-1)
+
+      gathered = tf.gather_nd(data_swaped, index_expanded)
+      y = tf.transpose(gathered, perm=axis_perm)
```
chinhuang007, line 54: The code looks great. Just a suggestion, could we provide a little performance enhancement by reducing transpose when axis is 0? The idea is to minimize the size of the graph as well as the expensive operations like transpose.

pluradj: updated, please review again

Issue #552, opened 2020-02-03
Dropout op should be enabled


winston-zillow:
The `Dropout` op is disabled due to these lines:
```python
    if cls.SINCE_VERSION >= 7 or attrs.pop("is_test", 0) == 1:
      return [x]
```
It appears a deliberated decision, perhaps because ONNX itself doesn't know what to do about the drop rate handling? This is a major blunder for trying to continuously train/fine-tune on existing imported model.

I suggest that `onnx-tensorflow` automatically generated a placeholder holding a boolean indicating whether it is in training and then it compute a dropout rate tensor based on the placeholder value. Here are the code skeleton inspired by [this post](https://stackoverflow.com/questions/44971349/how-to-turn-off-dropout-for-testing-in-tensorflow/44971517)

First add a command line option: `is_training_tensor_name`, default to `is_training`.

```python
class BackendHandler(Handler):
  @classmethod
  def get_or_create_tf_is_training(cls):
    is_training_tensor = command line option is_training_tensor_name
    with tf.name_scope(""):
      try:
        tf_is_training_float = tf.get_default_graph().get_tensor_by_name('%s_float:0' % is_training_tensor)
      except KeyError:
        tf_is_training = tf.placeholder_with_default(False, tuple(), name=is_training_tensor)
        tf_is_training_float = tf.cast(tf_is_training, tf.float32, name='%s_float' % is_training_tensor)
      return tf_is_training_float

class Dropout(BackendHandler):
  @classmethod
  def _common(cls, node, **kwargs):
    x = kwargs["tensor_dict"][node.inputs[0]]
    attrs = copy.deepcopy(node.attrs)
    # by default generate a placeholder with default value set to False, 
    # can skip the node by command-line option for small optimization
    if attrs.pop("is_test", 0) == 1 or option is_training_tensor_name is None
       return [x]
    rate = attrs.pop("ratio", 0.5)
    tf_is_training = cls.get_or_create_tf_is_training()
    return [tf.nn.dropout(x, rate=tf.multiply(tf_is_training, rate))]
```

winston-zillow:
CC @winnietsang @chinhuang007

winnietsang:
@winston-zillow currently there is no use case for the above usage, because ONNX-TF doesn't support training. We can revisit this once training capability is added to ONNX-TF.

gautambathla:
[...]

PR #671, opened 2020-06-17
Updated Dropout operator to support Opset 12 and 13.

shahirdaya:
As of Opset 12, the Dropout operator now supports training mode. I have created tests for this. The test is an average of 20 runs of the operator. It checks that the ratio of dropouts in the output is almost equal to the  ratio requested.
Opset 13 supports a bfloat16 input tensor. I have verified that tf.nn.dropout works fine with bfloat16 tensors. I have not created a test case for this.
This will close issue #552


chinhuang007:
A few onnx backend tests failed, as seen in CI. You can run `python test_onnx_backend.py -v OnnxBackendNodeModelTest.test_training_dropout_cpu` to see one of them. I am not sure the tests should produce the exact same results. Please take a look. Thanks!

shahirdaya:
@chinhuang007 I am looking into this. Thanks

chinhuang007 (approved):
LGTM!

Issue #637, opened 2020-06-02
Add latest pooling ops to the tf-1.x branch


sdmonov:
Few major issues were fixed in the latest pooling ops (Max and Avg) and will be good to have them in the tf-1.x branch

PR #674, opened 2020-06-18
Backport latest pooling ops into tf-1.x branch

sdmonov:
This PR backports latest max and average pool ops into tf-1.x branch.
Deprecated warnings can be seen from tensorflow about tf.nn.max/avg_pool_v2. The proposed replacement tf.nn.max/avg_pool does not fully support N-D tensors in tf 1.x and cannot be used directly. As this code is intended for tf 1.x use only this should be ok.

Fixed: #637


sdmonov:
CC @chinhuang007 @winnietsang

sdmonov:
@winnietsang @chinhuang007  added support for lppooling into tf-1.x. can you take a look? thx

winnietsang:
LGTM!

PR #685, opened 2020-06-25
LpPool operator implementation

sdmonov:
Implements LpPool by utilizing avg_pool:

    window_size = np.prod(ksize)

    input = tf.math.pow(input, self.p) * window_size
    pooled = tf.nn.avg_pool(input, ksize=ksize, strides=strides, padding=padding)
    pooled = tf.math.pow(pooled, 1.0 / self.p)

### DIFF
--- a/test/backend/test_node.py
+++ b/test/backend/test_node.py
@@ -1411,10 +1411,10 @@ class TestNode(unittest.TestCase):
                     count_include_pad=None,
                     pooling_type="MAX",
                     input_dtype=np.float32,
-                    p=2):
+                    p=None):
 
     op = "MaxPool" if pooling_type.upper().startswith("MAX") else \
-         "AveragePool" if pooling_type.upper()=="AVG" else "LpPool"
+         "AveragePool" if pooling_type.upper() == "AVG" else "LpPool"
     node_def_kwargs = {
         "op_type": op,
         "inputs": ["X"],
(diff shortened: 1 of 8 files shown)


test/backend/test_node.py
```diff
@@ -1899,6 +1902,102 @@ def test_average_pool_3d(self):
...
+  def test_lp2_pool_2d(self):
+    kernel_shape = [1, 2]
+    strides = [1, 2]
+    p = 2
+
+    input_shape = [10, 10, 4, 4]
+    self._test_pooling(input_shape=input_shape,
+                       kernel_shape=kernel_shape,
+                       strides=strides,
+                       p=p)
```
chinhuang007, line 1914: Looks like pooling_type is missing so it will used the default "MaxPool". Same for other lp test cases.

test/backend/test_node.py
```diff
@@ -1446,9 +1446,11 @@ def _test_pooling(self,
...
-                    input_dtype=np.float32):
+                    input_dtype=np.float32,
+                    p=2):
 
-    op = "MaxPool" if pooling_type.upper().startswith("MAX") else "AveragePool"
+    op = "MaxPool" if pooling_type.upper().startswith("MAX") else \
+         "AveragePool" if pooling_type.upper()=="AVG" else "LpPool"
     node_def_kwargs = {
         "op_type": op,
         "inputs": ["X"],
```
chinhuang007, line 1456: We probably need to add p to node_def_kwargs too to test lp_pool?

test/backend/test_node.py
sdmonov: Thanks for catching this. Fixed in the new commit.

test/backend/test_node.py
sdmonov: Thanks for catching this. Fixed in the new commit.

PR #706, opened 2020-07-17
Add auto data type cast

chinhuang007:
Add auto data type cast to support the data types that are
not natively supported by Tensorflow.

This PR adds the following
1. a common sys_config so the auto_cast option can be accessed
everywhere
2. additional data type support for equal operator
3. additional data type support for cumsum operator
4. unit tests for these data types
5. a new parameter auto-cast for CLI and prepare

### DIFF
--- a/onnx_tf/handlers/backend/cumsum.py
+++ b/onnx_tf/handlers/backend/cumsum.py
@@ -1,40 +1,45 @@
 import tensorflow as tf
 
-from onnx_tf.common import exception
 from onnx_tf.handlers.backend_handler import BackendHandler
 from onnx_tf.handlers.handler import onnx_op
-from onnx_tf.handlers.handler import partial_support
-from onnx_tf.handlers.handler import ps_description
 from onnx_tf.handlers.handler import tf_func
+from onnx_tf.common import sys_config
+from onnx_tf.common import exception
+from onnx_tf.common import data_type
+#import onnx_tf.common.data_type as data_type
 
 
 @onnx_op("CumSum")
 @tf_func(tf.math.cumsum)
-@partial_support(True)
-@ps_description(
-    "CumSum inputs in uint32/uint64 " + "are not supported in Tensorflow."
-)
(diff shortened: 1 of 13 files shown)


onnx_tf/handlers/backend/cumsum.py
```diff
@@ -1,35 +1,32 @@
...
+      x = tf.cast(x, cast_map[x.dtype]) if x.dtype in cast_map else x
+      return x
+
     tensor_dict = kwargs["tensor_dict"]
     x = tensor_dict[node.inputs[0]]
+
+    # handle data types that are not natively supported by Tensorflow
+    dtype = x.dtype
+    need_cast = dtype in [tf.uint32, tf.uint64]
+    x = dtype_cast(x) if need_cast else x
```
winnietsang, line 29: @chinhuang007 in the case that user set auto_cast=False and x.dtype=tf.uint64, then this cumsum op will still send to TF and TF will throw exception about this unsupported datatype. 
How about throw this exception in args_check and stop the conversion, instead of waiting for TF to throw this exception?

chinhuang007: Thanks for the comments. I added args_check for both equal and cumsum so exception will be thrown from our code. Please review again.

chinhuang007:
@winnietsang this is ready for review again. Thanks!

chinhuang007:
@winnietsang Let's change the python class style in a separate PR. I would prefer to leave it as is to be consistent with other exception classes.

test/backend/test_node.py
```diff
@@ -18,6 +18,7 @@
 from onnx_tf.common import supports_device
 from onnx_tf.common.legacy import legacy_onnx_pre_ver, legacy_opset_pre_ver
 from onnx_tf.common.pooling_helper import py_pool
+from onnx_tf.common import sys_config
```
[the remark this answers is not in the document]
chinhuang007: Thanks for the review. Line removed and typo fixed.

PR #714, opened 2020-07-23
Fix Travis build

pluradj:
* Include targets against ONNX official releases
* Allow failures against ONNX latest dev branch


test/backend/test_onnx_backend.py
```diff
@@ -132,6 +132,22 @@ def skip_not_implemented_ops_test(test):
...
+# LessOrEqual
+backend_test.exclude(r'test_less_equal_[a-z,_]*')
+# NegativeLogLikelihoodLoss and SoftmaxCrossEntropyLoss
+backend_test.exclude(r'test_softmax_cross_entropy_[a-z,_]*')
+# Dropout version 12
+backend_test.exclude(r'test_training_dropout_[a-z,_]*')
+# fails tolerance
+backend_test.exclude(r'test_gru_seq_length_[a-z,_]*')
+# KeyError: ''
+backend_test.exclude(r'test_nesterov_momentum_[a-z,_]*')
```
chinhuang007, line 149: The filtering here currently is purposed to skip the tests we know onnx-tf will not pass for known reasons and we have not plans in the future to enable the tests. It is not targeted for any particular ONNX opset version so a developer can run and see the failed cases against any ONNX version, such as 1.6, 1.7, or master.

Is it possible to separate the filtering for the tests that we wanted to skip all the time from the ones we plan to implement but not getting to them yet?

Also, if the goal is to pass CI, especially for the target ONNX version, maybe we need to introduce the ONNX version (opset version) to the code?

chinhuang007:
@pluradj This looks really good. One more request, could you please also remove the files, setup_onnx_master.sh and setup_onnx_target.sh, that are no longer needed with this approach?
Since we are not using docker for CI anymore, I think we should clean up the inactive files to make our repo clean. Thanks!

chinhuang007:
@pluradj Just saw a typo in https://github.com/onnx/onnx-tensorflow/blob/36a74f56f1f9dcd152bdb9319184f14c7551f179/onnx_tf/common/data_type.py#L104, unit64 should be uint64, caught by the CI. Could you please make the typo fix as well? Thanks!

pluradj:
> @pluradj Just saw a typo in
> 
> https://github.com/onnx/onnx-tensorflow/blob/36a74f56f1f9dcd152bdb9319184f14c7551f179/onnx_tf/common/data_type.py#L104
> , unit64 should be uint64, caught by the CI. Could you please make the typo fix as well? Thanks!

Updated, removed Docker pieces from Travis, and also added some backend exclusions for `pow` operator.

chinhuang007:
Thanks a lot for getting CI fixed!

Issue #710, opened 2020-07-21
ValueError for OneHot when attempting to prepare a TensorFlow representation


rallen10:
**Describe the bug**

I am trying to run inference on an onnx model with TensorFlow, but when I attempt to prepare the TF representation I receive a ValueError pertaining to a OneHot node as follows:

```
ValueError: Input must be scalar but has rank 1 for '{{node one_hot}} = OneHot[T=DT_FLOAT, TI=DT_INT32, axis=-1](strided_slice__20, const_fold_opt__56, strided_slice_3, strided_slice_2)' with input shapes: [2147483647], [1], [], [] and with computed input tensors: input[1] = <2>.
```

**To Reproduce**

```
import onnx 
from onnx_tf.backend import prepare
onnx_model = onnx.load('model.onnx')
tf_model = prepare(onnx_model)
```

**ONNX model file**

https://drive.google.com/file/d/1x_yli6xYUB4sa-YWlj90zIvBbiW0butF/view?usp=sharing

**Python, ONNX, ONNX-TF, Tensorflow version**

 - Python version: 3.7.6
 - ONNX version: 1.7.0
 - ONNX-TF version: git+https://github.com/onnx/onnx-tensorflow.git@c60a32caef3271b93e843bac7a44eda388f67165
 - Tensorflow version: 2.2.0

**Additional context**

As best as I can tell, the error stems from the fact that the onnx model stores the `depth` value of the one_hot function as a rank 1 tensor instead of as a scalar. I did not generate the onnx model myself, but I know it was generated in Unity ML-Agents using tf2onnx

chinhuang007:
Thanks for opening the issue. The problem is coming from ONNX doc, stating depth is a scalar, https://github.com/onnx/onnx/blob/master/docs/Operators.md#onehot. Tensorflow throws the exception because the depth in your model is a 1D tensor of size 1. However the spec says depth could be a single-element vector, https://github.com/onnx/onnx/blob/21b2cda6066ca4b37ba9b1fec915d5dca9cc12af/onnx/defs/tensor/defs.cc#L2340. This should be fixed to be consistent in ONNX. For now, we could look into a fix to support 1D tensor.

PR #718, opened 2020-07-27
Update OneHot to support more use cases

chinhuang007:
Update OneHot to support the following use cases:
1. Data types that are not natively supported by Tensorflow
2. Negative indices
3. Negative axis
4. Depth as 1D tensor of size 1 (issue
https://github.com/onnx/onnx-tensorflow/issues/710)

The PR also removes the filter on onnx backend tests for OneHot.


test/backend/test_onnx_backend.py
```diff
@@ -85,7 +85,7 @@ def skip_not_implemented_ops_test(test):
 
 # TF only support uint8, int32, int64 for indices and int32 for depth in
 # tf.one_hot
-backend_test.exclude(r'test_onehot_[a-z,_]*')
+#backend_test.exclude(r'test_onehot_[a-z,_]*')
```
winnietsang, line 88: @chinhuang007 please remove this line.

PR #722, opened 2020-07-29
Moved onnx imports before tensorflow

### LINKED ISSUE #721: Some tests fail to run on Mac against ONNX 1.7.0
**Describe the bug**

I attempted to run `python test/backend/test_node.py` and encountered this error:

```
[libprotobuf ERROR google/protobuf/descriptor_database.cc:394] Invalid file descriptor data passed to EncodedDescriptorDatabase::Add().
[libprotobuf FATAL google/protobuf/descriptor.cc:1356] CHECK failed: GeneratedDatabase()->Add(encoded_file_descriptor, size):
libc++abi.dylib: terminating with uncaught exception of type google::protobuf::FatalException: CHECK failed: GeneratedDatabase()->Add(encoded_file_descriptor, size):
Abort trap: 6
```

This error does not appear when using ONNX 1.6.0 on Mac. Not an issue on Ubuntu 18.04 or Windows Server 2016 either.

**To Reproduce**

Running on macOS Catalina Version 10.15.6 with Python 3.6.10

```
pip install onnx==1.7.0 tensorflow==2.2.0 tensorflow-addons==0.10.0 pytest tabulate
git clone https://github.com/onnx/onnx-tensorflow.git
cd onnx-tensorflow/
git checkout rel-1.6.0
pip install -e .
python test/backend/test_node.py TestNode.test_add
```

**ONNX model file**

N/A

**Python, ONNX, ONNX-TF, Tensorflow version**

Python version:
3.6.10 (default, Jul  9 2020, 23:47:59)
[GCC 4.2.1 Compatible Apple LLVM 11.0.3 (clang-1103.0.32.62)]
ONNX version:
1.7.0
ONNX-TF version:
1.6.0
Tensorflow version:
2.2.0

**Additional context**

This is similar to issue reported on tf2onnx https://github.com/onnx/tensorflow-onnx/issues/1002. The fix there was to `import onnx` before tensorflow related imports. I'm not clear on what the underlying cause is, but I have a pull request prepared for a similar fix in `test_node.py` and `test_dynamic_shape.py` which allows these tests to run. `test_cli.py`, `test_model.py`, and `test_onnx_backend.py` were able to run without any changes.

pluradj:
* Fixes #721

### DIFF
--- a/test/backend/test_dynamic_shape.py
+++ b/test/backend/test_dynamic_shape.py
@@ -3,15 +3,17 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import numpy as np
 import unittest
+
+from onnx import defs
+from onnx import helper
+from onnx import TensorProto
+import numpy as np
+import tensorflow as tf
 
 from onnx_tf.backend import onnx_graph_to_tensorflow_rep
 from onnx_tf.common.legacy import legacy_opset_pre_ver
 from onnx_tf.common.pooling_helper import py_pool
-from onnx import defs
-from onnx import helper
-from onnx import TensorProto
 
 
 class TestDynamicShape(unittest.TestCase):
--- a/test/backend/test_node.py
+++ b/test/backend/test_node.py
@@ -6,16 +6,18 @@ from __future__ import unicode_literals
 import sys
 import math
 import unittest
+
+from onnx import helper
+from onnx import TensorProto
+from onnx import defs
 import numpy as np
 import tensorflow as tf
+
 from onnx_tf.backend import onnx_graph_to_tensorflow_rep
 from onnx_tf.backend import run_node
 from onnx_tf.common import supports_device
 from onnx_tf.common.legacy import legacy_onnx_pre_ver, legacy_opset_pre_ver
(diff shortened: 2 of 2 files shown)


test/backend/test_dynamic_shape.py
```diff
@@ -3,17 +3,17 @@
...
 from onnx_tf.common.legacy import legacy_opset_pre_ver
 from onnx_tf.common.pooling_helper import py_pool
 from onnx import defs
 from onnx import helper
 from onnx import TensorProto
 
+import numpy as np
+import tensorflow as tf
+import unittest
+
```
chinhuang007, line 16: We have recommended to follow the import order, with 3 main sections:
1. system imports
2. software dependency imports, ie onnx and tensorflow
3. own packages, ie onnx_tf

Examples can be found in backend.py, converter.py, etc.

As you found here, not all files follow the recommend import order. I would like suggest we keep the same 3 sections whenever possible. So if needed, we just need to ensure onnx import comes before tensorflow in the middle sections.

chinhuang007: The changes look good! Can you please resolve the branch conflicts?

PR #725, opened 2020-08-05
Add Pow support for opset 12 and 13

chinhuang007:
Add Pow support for opset 12 and 13 including
1. add data type cast whenever needed
2. add support when input and exponent have different types
3. add test case for bfloat16 in test_model.py

Also updated .gitignore so the generated version.py is not
tracked by git

### DIFF
--- a/onnx_tf/handlers/backend/pow.py
+++ b/onnx_tf/handlers/backend/pow.py
@@ -4,6 +4,8 @@ from onnx_tf.handlers.backend_handler import BackendHandler
 from onnx_tf.handlers.handler import onnx_op
 from onnx_tf.handlers.handler import tf_func
 from .math_mixin import BasicMathMixin
+from onnx_tf.common import sys_config
+from onnx_tf.common import data_type
 
 
 @onnx_op("Pow")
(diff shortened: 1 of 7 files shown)


onnx_tf/handlers/backend/pow.py
```diff
@@ -4,16 +4,75 @@
...
+          tf.int8: [tf.float16, tf.float32, tf.float64, tf.int32, tf.int64],
+          tf.int16: [tf.float16, tf.float32, tf.float64, tf.int32, tf.int64],
+          tf.int32: [tf.float32, tf.float64, tf.int64],
+          tf.int64: [tf.float64],
+          tf.uint8: [tf.float16, tf.float32, tf.float64, tf.int32, tf.int64],
+          tf.uint16: [tf.float32, tf.float64, tf.int32, tf.int64],
+          tf.uint32: [tf.float64, tf.int64]
+      }
+      return to_dtype in safe_cast_map[from_dtype]
+
```
winnietsang, line 31: @chinhuang007 may be move this "is_safe_cast" function to onnx_tf.common so other handler can use this function too.

chinhuang007: Good idea. Moved "is_safe_cast" to common.data_type.py and added complex types. Please review. Thanks!

Issue #547, opened 2020-01-24
onnx to tensorflow conversion error: "module 'tensorflow' has no attribute 'ceil'"


TheArheus:
**Describe the bug**

Hi, I'm trying to import my onnx model, which is exported from PyTorch. But when I tried to to use `prepare()` method, first of all, I started to get errors like this:

```
C:\Users\1\Anaconda3\lib\site-packages\onnx_tf\common\__init__.py:89: UserWarning: onnx_tf.common.get_outputs_names is deprecated. It will be removed in future release. Use TensorflowGraph.get_outputs_names instead.
  warnings.warn(message)
Traceback (most recent call last):
  File "C:/Users/1/PycharmProjects/untitled/onnx_.py", line 3, in <module>
    from onnx_tf.backend import prepare
  File "C:\Users\1\Anaconda3\lib\site-packages\onnx_tf\__init__.py", line 1, in <module>
    from . import backend
  File "C:\Users\1\Anaconda3\lib\site-packages\onnx_tf\backend.py", line 29, in <module>
    from onnx_tf.common.handler_helper import get_all_backend_handlers
  File "C:\Users\1\Anaconda3\lib\site-packages\onnx_tf\common\handler_helper.py", line 5, in <module>
    from onnx_tf.handlers.backend import *  # noqa
  File "C:\Users\1\Anaconda3\lib\site-packages\onnx_tf\handlers\backend\ceil.py", line 10, in <module>
    @tf_func(tf.ceil)
AttributeError: module 'tensorflow' has no attribute 'ceil'
```

But after trying to fix all problems like this I got next error:
```
C:\Users\1\Anaconda3\lib\site-packages\onnx_tf\common\__init__.py:89: UserWarning: onnx_tf.common.get_outputs_names is deprecated. It will be removed in future release. Use TensorflowGraph.get_outputs_names instead.
  warnings.warn(message)
Traceback (most recent call last):
  File "C:/Users/1/PycharmProjects/untitled/onnx_.py", line 17, in <module>
    tf_rep = prepare(model)
  File "C:\Users\1\Anaconda3\lib\site-packages\onnx_tf\backend.py", line 65, in prepare
    return cls.onnx_model_to_tensorflow_rep(model, strict)
  File "C:\Users\1\Anaconda3\lib\site-packages\onnx_tf\backend.py", line 85, in onnx_model_to_tensorflow_rep
[...]

PR #733, opened 2020-08-15
Update README.md

usamec:
Per: https://github.com/onnx/onnx-tensorflow/issues/547#issuecomment-640720149


chinhuang007:
Thanks for reporting the issue. The pip install is supposed to work for each release, including 1.6. The problem is tracked https://github.com/onnx/onnx-tensorflow/issues/738. It should be addressed shortly.

chinhuang007:
The pip install is working now; therefore close the PR.

PR #757, opened 2020-09-08
Implemented Opset 13 for Add, Ceil, Erf, Exp, and Floor operators.

shahirdaya:
No additional tests were required as in all of these operators, the only change is support for bfloat16 which all of the corresponding TF operators support.

Add: from Opset 7 to 13 only difference is bfloat16 support. 
Ceil: from Opset 6 to 13 only difference is bfloat16 support. 
Erf: from Opset 9 to 13 only difference is bfloat16 support. 
Exp: from Opset 6 to 13 only difference is bfloat16 support. 
Floor: from Opset 6 to 13 only difference is bfloat16 support.


chinhuang007 (approved):
Looks good! Thank you!

PR #543, opened 2020-01-23
fix undefined shape/batch size for slice operator

winston-zillow:
The `slice` operator conversion failed to handle tensor shape with undefined/batch size dimension. This PR fixes it by querying the dynamic shape instead of creating a constant from static shape.

PR #771, opened 2020-09-26
use tf.shape instead of .shape for dynamic axes

masakistan:
`InstanceNormalization ` failed to handle tensor shape with undefined/batch size dimension. This PR fixes it by querying the dynamic shape instead of creating a constant from static shape.

Essentially the same as #543


onnx_tf/handlers/backend/instance_normalization.py
```diff
@@ -31,7 +31,7 @@ def _common(cls, node, **kwargs):
     beta = tensor_dict[node.inputs[2]]
 
     inputs = tensor_dict[node.inputs[0]]
-    inputs_shape = inputs.shape
+    inputs_shape = tf.shape(inputs)
```
winnietsang, line 34: We have a common util function call "tf_shape" in onnx-tensorflow/onnx_tf/common/tf_helper.py that would check the shape of the input is fully defined or not to decide to use numpy or tensorflow op. May you please use our "tf_shape" instead of "tf.shape".

winnietsang (changes requested):
@masakistan Thanks for helping us to identify and fix this dynamic shape issue. 
May you please add in a dynamic shape testcase for instance_normalization into onnx-tensorflow/test/backend/test_dynamic_shape.py

onnx_tf/handlers/backend/instance_normalization.py
masakistan: Thanks! I've made the suggested changes. Please let me know if anything else needs to change.

winnietsang:
@masakistan please add in a dynamic shape testcase for instance_normalization into onnx-tensorflow/test/backend/test_dynamic_shape.py

masakistan:
> @masakistan please add in a dynamic shape testcase for instance_normalization into onnx-tensorflow/test/backend/test_dynamic_shape.py

Will do, sorry I haven't had time to get to it.

mcourse1993:
> @masakistan please add in a dynamic shape testcase for instance_normalization into onnx-tensorflow/test/backend/test_dynamic_shape.py

Hi, I want to convert my onnx file to the tensorflow considering dynamic batch size, but it fails after lots of trial, could you help me
I've attached the onnx file here
https://drive.google.com/file/d/15m7fXX7Q2M5EMVLxbiqeUktKbjZPdlST/view?usp=sharing

PR #775, opened 2020-10-01
Add data cast support for Clip and Mod

chinhuang007:
Add data cast support for Clip and Mod so that all ONNX data
types are supported for these two operators now.

Also add a test case in test_node.py for mod to verify data
cast working.


winnietsang (approved):
LGTM

PR #798, opened 2020-11-05
Fix lstm export to saved model issue

chinhuang007:
Fix lstm export to saved model issue due to the use of
tf.Variable in the handler. Also add a test case to
test_model.py to verify a model can be exported and loaded
back successfully.


winnietsang (approved):
LGTM!

PR #800, opened 2020-11-10
Update package dependency versions for TravisCI

pluradj:


chinhuang007 (approved):
Looks good. Thanks for the clarification as well.

PR #789, opened 2020-10-20
Add RoiAlign support

grimoire:
Hi
This PR add RoiAlign support to tf-1.x.
`sampling_ratio`>0 part are fine. But I didn't manage to fully implement `sampling_ratio`<=0 case with python only.
The results of `sampling_ratio`<=0 might be **flatten** than other implementation(onnxruntime/pytorch)

PR #805, opened 2020-11-13
add roi_align support to tf2.x

grimoire:
Hi
This repo add roi_align support to TF2.x, same logic as https://github.com/onnx/onnx-tensorflow/pull/789
I am new to TF2.x, hope this works fine.


chinhuang007 (approved):
I ran the onnx standard test for RoiAlign with TF 2.3.1. It looks good!

PR #806, opened 2020-11-15
Updated Hardmax for opset 13 

shahirdaya:
The only change from opset 11 to 13 was support bfloat16 and also the default for axis is -1 in opset 13 as opposed to 1 in opset 11.

It appears that the implementation of converting to a 2d, doing a hardmax, and reshaping the
result did not really work. I  have changed the implementation to be transpose based and I
flip the result.

All tests pass.

Signed-off-by: Shahir A. Daya <sdaya@ca.ibm.com>

### DIFF
--- a/test/backend/test_node.py
+++ b/test/backend/test_node.py
@@ -9,8 +9,10 @@ import unittest
 from onnx import defs
 from onnx import helper
 from onnx import TensorProto
+from onnx.backend.test.case.node import hardmax
 import numpy as np
 import tensorflow as tf
+import tensorflow_addons as tfa
 
 from onnx_tf.backend import onnx_graph_to_tensorflow_rep
 from onnx_tf.backend import run_node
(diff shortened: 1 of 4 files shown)


test/backend/test_node.py
```diff
@@ -1056,11 +1057,23 @@ def test_hardmax(self):
...
+      axis = axis if axis >= 0 else len(np.shape(x)) + axis
+      if axis == len(np.shape(x)) - 1:
+          np.testing.assert_almost_equal(output["Y"], tfa.seq2seq.hardmax(x))
+      else:
+          if not legacy_opset_pre_ver(13):
+            perm1 = tf.range(0, axis)
+            perm2 = tf.range(axis + 1, len(tf.shape(x)) - 1)
+            perm = tf.concat([perm1, [len(tf.shape(x)) - 1], perm2, [axis]], -1)
+            x_transposed = tf.transpose(x, perm)
+            np.testing.assert_almost_equal(output["Y"], tf.transpose(tfa.seq2seq.hardmax(x_transposed), perm))
```
winnietsang, line 1070: @shahirdaya how about using hardmax defined in onnx.backend.test.case.node.hardmax to calculate the expected value here, then it will be using different way to calculate the expected value and compare the value return from the hardmax handler.

winnietsang (changes requested):

test/backend/test_node.py
shahirdaya: Thanks @winnietsang. Done.

winnietsang (approved):
LGTM

PR #815, opened 2020-11-20
Fix auto cast for data type downcast

chinhuang007:
Fix auto cast for data type downcast by moving the auto-cast
check to args_check. Also add a simple test to verify the user
argument is applied for downcast when needed.


winnietsang (approved):
LGTM

PR #827, opened 2020-12-02
Fix auto cast checking for clip and sign

chinhuang007:
Fix auto cast checking for clip and sign handlers so
the user option will be properly applied


winnietsang (approved):
LGTM

PR #838, opened 2020-12-15
Fix Slice version 1 to support model zoo models

chinhuang007:
Fix Slice version 1 to support model zoo models:
age_gender, version-RFB-320.onnx
age_gender, version-RFB-640.onnx


winnietsang (approved):
LGTM!

PR #839, opened 2020-12-16
Add data cast to fix LessOrEqual issue with tf 2.4

chinhuang007:
Add data cast to fix LessOrEqual issue with tf 2.4, since
tf LessEqual doesn't support uint16, uint32, uint64. Also
add test code to verify unsupported data types.


winnietsang (approved):
LGTM!

PR #861, opened 2021-01-28
Add new data types support for Pad

chinhuang007:
Add new data types support for Pad, which gets
string, bool, and bfloat16. Also add a test case
for string type.


winnietsang (approved):
LGTM

Issue #470, opened 2019-07-25
Depth Separable Convolution


JoseJoji:
I am converting a Matlab model to ONNX and then to Tensorflow. It has a depth separable convolution node in my net. Everything works fine with the latest version of ONNX and ONNX-TF and TF version 1.14. But the depth separable convolution node is inferred as multiple convolution kernels. Do the latest ONNX-TF support Depth Separable Convolution ?

maxim-gusarov:
Same here
<img width="941" alt="Screen Shot 2019-07-25 at 1 58 38 PM" src="https://user-images.githubusercontent.com/51131273/61908358-55fe0c80-aee4-11e9-9948-dd0593f6973e.png">

ainrichman:
PyTorch and other frameworks does not support native depthwiseconv op. Could onnx-tf give a logic to determine if the groupped conv is depthwise or not?

tigert1998:
In 2020, onnx-tf still cannot detect dwconv automatically

anuragmundhada:
What would be a workaround for this? I need to convert a pytorch model to tensorflow, the graph creates tons of regular convolution operations instead of a depthwise separable convolution.

chinhuang007:
Can you provide the onnx model file? Onnx-tf currently doesn't do node fusion. Your contribution is more than welcome.

anuragmundhada:
> Can you provide the onnx model file? Onnx-tf currently doesn't do node fusion. Your contribution is more than welcome.

The attached ONNX model still has 69 convolution layers, same as my Pytorch model. The frozen graph after tf_rep.export_graph contains 1583 convolution layers.

I am new to this repo but would like to contribute. Will look into the code.

[model.zip](https://github.com/onnx/onnx-tensorflow/files/4638086/model.zip)

zetyquickly:
Is someone working on this issue?

Elecky:
I also encountered this problem, and implemented a rough solution.: https://github.com/Elecky/onnx-tensorflow/tree/dev/depthwise_conv (based on tf-1.x branch)
I'd like to contribute a PR if necessary!

InCogNiTo124:
@CyanHillFox your code looks like something I'd need but I cannot run it because it is not tf2. Did you plan to ever upgrade that to tf2?

CyanHillFox:
> @CyanHillFox your code looks like something I'd need but I cannot run it because it is not tf2. Did you plan to ever upgrade that to tf2?

[...]

PR #867, opened 2021-02-08
Fix Conv2D for case of group for DepthwiseConv

seanshpark:
This will try to convert to DepthwiseConv with condition of group value.
- based on https://github.com/onnx/onnx-tensorflow/issues/470#issuecomment-681584457 by CyanHillFox@github

Signed-off-by: SaeHie Park <saehie.park@samsung.com>

### DIFF
--- a/onnx_tf/handlers/backend/conv_mixin.py
+++ b/onnx_tf/handlers/backend/conv_mixin.py
@@ -91,16 +91,45 @@ class ConvMixin(BroadcastMixin):
                                         "Tensorflow")
 
     group = node.attrs.get("group", 1)
+    weight_shape = weights.get_shape().as_list()
+    # Is this convolution depthwise we can support?
+    depthwise = (len(x_shape) == 4 and len(weight_shape) == 4 and
+                group != 1 and group == x_shape[1] and not transpose and
+                not (None in weight_shape))
 
-    weight_groups = tf.split(weights, num_or_size_splits=group, axis=-1)
+    if depthwise is True:
+      # Depthwise convolution.
+      # The convolution kernel layout in tf.depthwise_conv is:
+      # [filter_height, filter_width, in_channels, channel_multiplier]
+      # Weight is now (KH x KW X C/g X M), or more precisely, (KH x KW X C/g X (g * M/g)),
+      # we reshape it to (KH x KW x C x M/g)
+      # NOTE: Assuming weight has fixed shape.
 
-    if sys_config.device == 'CUDA':
-      xs = tf.split(x, num_or_size_splits=group, axis=1)
+      depthwise_filter_shape = weight_shape[0:2] + [-1, weight_shape[3] // group]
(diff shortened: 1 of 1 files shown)


seanshpark:
I've go this error in one of the CI but I don't know how to reproduce this... can anyone help how to reproduce or how to fix?
```
ERROR: test_identity_sequence_cpu (backend.test_onnx_backend.OnnxBackendNodeModelTest)
```

chinhuang007:
@seanshpark Thanks for the contribution! The CI problem is introduced by a recently merged PR in ONNX. I already created a patch, https://github.com/onnx/onnx-tensorflow/pull/868. So please don't worry about it.

onnx_tf/handlers/backend/conv_mixin.py
```diff
@@ -91,16 +91,47 @@ def conv(cls, node, input_dict, transpose=False):
...
-    weight_groups = tf.split(weights, num_or_size_splits=group, axis=-1)
+    if depthwise is True:
+      # Depthwise convolution.
+      # The convolution kernel layout in tf.depthwise_conv is:
+      # [filter_height, filter_width, in_channels, channel_multiplier]
+      # Weight is now (KH x KW X C/g X M), or more precisely, (KH x KW X C/g X (g * M/g)),
+      # we reshape it to (KH x KW x C x M/g)
+      # NOTE: Assuming weight has fixed shape.
+      if weight_shape is None or None in weight_shape or group is None:
+        depthwise = False
```
chinhuang007, line 107: Is it possible to combine code to determine whether depthwise in True? For example, if len(weight_shape) == 4, then weight_shape is not None. And looks like the logic for group could be simplified too.

seanshpark: Thanks for the review! I'll revise the change :)

seanshpark:
@chinhuang007 , I'm not that good at python so better suggestion are welcome :)

seanshpark:
If the CI fails, I'll rebase after #868 lands

onnx_tf/handlers/backend/conv_mixin.py
```diff
@@ -224,16 +253,35 @@ def conv(cls, node, input_dict, transpose=False):
...
-          for (x, weight) in zip(xs, weight_groups)
-      ]
+    else: # not transpose:
+      if depthwise is True:
+        if compute_format == "NHWC":
+          strides = [1] + strides + [1]
+        elif compute_format == 'NCHW':
+          strides = [1, 1] + strides
+        else:
+          raise ValueError("Invalid storage_format: {}".format(storage_format))
```
chinhuang007, line 263: Should it be invalid compute_format?

seanshpark: Thanks!

chinhuang007 (approved):
LGMT!

PR #868, opened 2021-02-08
Fix Identity with sequence type

chinhuang007:
Fix Identity with sequence type with two workarounds since
Tensorflow Identity doesn't support sequence:
1. update identity handler as partial support
2. skip the standard ONNX test case for identity with sequence


winnietsang (approved):
LGTM

PR #871, opened 2021-02-16
Fix Conv input rank issue when shape is not fully defined

chinhuang007:
Fix the Conv input rank issue when the input shape is not fully
defined. The use of len(x_shape) would throw exception if the
x_shape is a tensor, returned by tf_shape function. The x_rank is
already captured and safe to use here.


winnietsang (approved):
LGTM

PR #879, opened 2021-03-13
Support training with batch normalization op; update training example

chudegao:
Update batch normalization  handler to support training 
Update training example:
  1. training tf model before save to to onnx model
  2. add vgg model to test larger mode performance
  3. remove optimizer op run when do test
  4. use all test data to run trained onnx model

### DIFF
--- a/onnx_tf/handlers/backend/batch_normalization.py
+++ b/onnx_tf/handlers/backend/batch_normalization.py
@@ -1,4 +1,5 @@
 import tensorflow as tf
+import onnx_tf.backend
 
 from onnx_tf.handlers.backend_handler import BackendHandler
 from onnx_tf.handlers.handler import onnx_op
@@ -27,8 +28,8 @@ class BatchNormalization(BackendHandler):
     x_shape = x.get_shape().as_list()
     x_rank = len(x_shape)
 
-    params_shape_broadcast = list(
-        [1, x_shape[1]] + [1 for _ in range(2, x_rank)])
+    params_shape_broadcast = list([1, x_shape[1]] +
+                                  [1 for _ in range(2, x_rank)])
     if params_shape_broadcast[1] is None:
       params_shape_broadcast[1] = tf.shape(x)[1]
       params_shape_broadcast = tf.stack(params_shape_broadcast)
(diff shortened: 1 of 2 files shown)


onnx_tf/handlers/backend/batch_normalization.py
```diff
@@ -27,33 +28,69 @@ def _common(cls, node, **kwargs):
...
-      return [cls.make_tensor_from_onnx_node(node, inputs=inputs)]
+    running_mean_1d = tensor_dict[node.inputs[3]]
+    running_var_1d = tensor_dict[node.inputs[4]]
+
+    # # from version 7, force to use test mode
+    # if cls.SINCE_VERSION >= 7 or node.attrs.get("is_test", 0):
+    #   inputs = [x, running_mean, running_variance, bias, scale]
+    #   return [cls.make_tensor_from_onnx_node(node, inputs=inputs)]
+
     spatial = node.attrs.get("spatial", 1) == 1
```
chinhuang007, line 49: Please keep the old code for SINCE_VERSION < 7. No need to enable training there, but we should still support older versions.

chudegao: The code commeted here is for SINCE_VERSION >= 7.  Do you mean we keep the code for  SINCE_VERSION < 7 like below?   Buts seems it's for training of SINCE_VERSION<7. 

if cls.SINCE_VERSION < 7 and not node.attrs.get("is_test", 0):
    spatial = node.attrs.get("spatial", 1) == 1
    momentum = node.attrs.get("momentum", 0.9)
    axis = [0] if spatial else [0] + list(range(2, total_num_dim))
    mean, variance = tf.nn.moments(x, axis)
    running_mean = running_mean * momentum + mean * (1 - momentum)
    running_variance = running_variance * momentum + variance * (1 - momentum)
    # TODO: need to conform to the documentation here
    inputs = [x, running_mean, running_variance, bias, scale]
    return [cls.make_tensor_from_onnx_node(node, inputs=inputs)]

chinhuang007: Yes, some code is commented out, and some code is removed. The problem is the new code here doesn't work for the SINCE_VERSION < 7 models. Please check https://github.com/onnx/onnx/blob/master/docs/Changelog.md#BatchNormalization-6. As you can see, the node has an attribute "is_test" to indicate the use is for training or not. We should not change the behavior or the original intent with our own flag. I would suggest to leave the old code there as is and just handle training for SINCE_VERSION >= 7.

chudegao: Updated, please help review.

chinhuang007 (approved):
LGTM!

PR #880, opened 2021-03-15
fix dynamic shape inference in DepthToSpace

d-miketa:
Nets using DepthToSpace are currently incompatible with dynamic axes. It's enough to switch from `x.shape` to `tf.shape(x)` along with manual unrolling of the resulting 4-dimensional tensor (as one is not allowed to directly iterate over a tf.Tensor).

Related to [#543](https://github.com/onnx/onnx-tensorflow/pull/543) and [#771](https://github.com/onnx/onnx-tensorflow/pull/771)

Edit: Added a unittest. It's my first one and my second PR ever, so fingers crossed..!

### DIFF
--- a/onnx_tf/handlers/backend/depth_to_space.py
+++ b/onnx_tf/handlers/backend/depth_to_space.py
@@ -3,6 +3,7 @@ import copy
 import tensorflow as tf
 
 from onnx_tf.common import get_data_format
+from onnx_tf.common.tf_helper import tf_shape
 from onnx_tf.handlers.backend_handler import BackendHandler
 from onnx_tf.handlers.handler import onnx_op
 from onnx_tf.handlers.handler import tf_func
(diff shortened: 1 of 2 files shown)


onnx_tf/handlers/backend/depth_to_space.py
```diff
@@ -40,7 +40,8 @@ def _common(cls, node, **kwargs):
     if mode == "CRD":
       # need native computation
       bsize = attrs.get("blocksize")
-      batch, channel, height, width = x.shape
+      x_shape = tf.shape(x)
```
chinhuang007, line 43: First, thanks very much for making the contribution! If you don't mind, please consider use the utility method tf_shape, which takes care of both fully defined and dynamic shapes, and provides slight optimization for static shape.

test/backend/test_dynamic_shape.py
```diff
@@ -248,6 +248,32 @@ def test_conv_transpose(self):
 
     np.testing.assert_almost_equal(tf_model_output[0], test_output, decimal=5)
 
+  def test_depth_to_space(self):
+    b,c,h,w = shape = [2, 48, 5, 6]
```
chinhuang007, line 252: The code looks fine. Please run yapf as described in https://github.com/onnx/onnx-tensorflow#code-standard to keep the code format and style consistent. Thanks.

onnx_tf/handlers/backend/depth_to_space.py
d-miketa: Thanks for the suggestion, switched to tf_shape.

test/backend/test_dynamic_shape.py
d-miketa: Thank you for pointing out that document, done on both files.

chinhuang007 (approved):
LGTM!

Issue #754, opened 2020-09-08
mobilenet_v2 in torchvision -> ONNX -> TensorFlow works but seems unexpected huge model


zlqhem:
**Reproduce**
details: https://github.com/zlqhem/example-pytorch2tf/blob/master/pytorch2tf.ipynb

1. pytorch -> ONNX
```
model = torchvision.models.mobilenet_v2(pretrained=True).cuda()
torch.onnx.export(model, dummy_input, MODEL_ONNX, verbose=True, input_names=input_names, output_names=output_names)
```
![onnx model](https://github.com/zlqhem/example-pytorch2tf/raw/master/mobilenet_v2.onnx.png)

2. ONNX -> tensorflow
```
onnx-tf convert -i mobilenet_v2.onnx -o ./mobilenet_v2.pb
```
it seems unexpected huge model.

![tensorflow model](https://github.com/zlqhem/example-pytorch2tf/raw/master/mobilenet_v2.pb.png)

**Python, ONNX, ONNX-TF, Tensorflow version**

Python: 3.7
ONNX: 1.7.0 
ONNX_TF: 1.6.0 (source build a few days ago)
TensorFlow: 2.3.0

zlqhem:
After following onnx-tensorflow, it seems reasonable for converting BatchNomalization node as like above picture. (4 Reshape, 1 Transponse and so on).
4 Reshape corresponds with mean, variance, bias, scale for handling batch normalization.

Question #1. Why not use BatchNomalization in tensorflow directly? Are there any issues? (Sorry, I'm new to both tensorflow and onnx-tensorflow)

But I wonder why converted Conv layer is so complex.

Below is the conv layer in onnx which is converted from mobilenet v2 in pytorch model. 
[](https://github.com/zlqhem/example-pytorch2tf/raw/master/conv-layer-in-onnx.png)

Below is the conv layer converted using onnx-tensorflow. 
[](https://github.com/zlqhem/example-pytorch2tf/raw/master/converted-conv-layer.1.png)
[](https://github.com/zlqhem/example-pytorch2tf/raw/master/converted-conv-layer.2.png)

If I'm something wrong, please let me know.

zlqhem:
After following onnx-tensorflow, it seems reasonable for converting BatchNomalization node as like above picture. (4 Reshape, 1 Transponse and so on).
4 Reshape corresponds with mean, variance, bias, scale for handling batch normalization.

[...]

PR #885, opened 2021-03-18
Fix depthwise condition check with unknown input shape

chinhuang007:
Fix depthwise condition check with unknown input shape for conv
where x_shape[1] could be a tensor and group == x_shape[1] will
be always False.

### DIFF
--- a/onnx_tf/handlers/backend/conv_mixin.py
+++ b/onnx_tf/handlers/backend/conv_mixin.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 import tensorflow as tf
 
 from onnx_tf.common import exception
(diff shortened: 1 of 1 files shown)


seanshpark:
@chinhuang007 , as wrote in https://github.com/onnx/onnx-tensorflow/issues/754#issuecomment-802342948, I'll try to find a fix as this change has problem converting to DWConv...

seanshpark (approved):
LGTM thank you!

onnx_tf/handlers/backend/conv_mixin.py
```diff
@@ -93,9 +95,14 @@ def conv(cls, node, input_dict, transpose=False):
     group = node.attrs.get("group", 1)
     weight_shape = weights.get_shape().as_list()
     # Is this convolution depthwise we can support?
-    depthwise = (x_rank == 4 and len(weight_shape) == 4 and
+    if isinstance(x_shape, np.ndarray):
+      depthwise = (x_rank == 4 and len(weight_shape) == 4 and
                 group != 1 and group == x_shape[1] and not transpose and
                 not (None in weight_shape))
+    else:
+      depthwise = (x_rank == 4 and len(weight_shape) == 4 and
```
Vozf, line 103: To avoid copying the full bool condition you may consider doing it this way
```
depthwise = (x_rank == 4 and len(weight_shape) == 4 and
                group != 1 and not transpose and
                not (None in weight_shape))
if isinstance(x_shape, np.ndarray):
    depthwise = (depthwise and group == x_shape[1])
```

chinhuang007: Code is simplified, same logic.

Issue #902, opened 2021-04-15
Typecast to bool missing for "depthwise" flag in onnx_tf/handlers/backend/conv_mixin.py


vijeth-kum:
When `x_shape` is of type np.ndarray, `depthwise` at line 101 is casted implicitly to np.bool and hence the check at line L103 fails.
```
L98    depthwise = (x_rank == 4 and len(weight_shape) == 4 and group != 1 and
L99                 not transpose and not (None in weight_shape))
L100    if depthwise and isinstance(x_shape, np.ndarray):
L101      depthwise = group == x_shape[1]
L102
L103    if depthwise is True:
```
**Expected**: `depthwise` at L101 should be casted to bool
depthwise = bool(group == x_shape[1])

chinhuang007:
Thanks for finding the bug. A fix will be provided shortly.

chinhuang007:
Fixed. Please reopen if not the case. Thanks.

PR #905, opened 2021-04-19
Fix depthwise condition bug

chinhuang007:
Fix depthwise condition bug, as described in issue https://github.com/onnx/onnx-tensorflow/issues/902


chinhuang007:
Thanks for your review!

PR #932, opened 2021-06-22
Add HardSwish support

chinhuang007:
Add HardSwish support, which is introduced in ONNX opset 14.
Add also a filter to exclude new non-deterministic ONNX model
tests.


onnx_tf/handlers/backend/hardswish.py
```diff
@@ -0,0 +1,20 @@
+import tensorflow as tf
+
+from onnx_tf.handlers.backend_handler import BackendHandler
+from onnx_tf.handlers.handler import onnx_op
+
+
+@onnx_op("HardSwish")
+class ThresholdedRelu(BackendHandler):
```
chudegao, line 8: The class name seems not right

chinhuang007: Thanks for catching this... fixed.

PR #938, opened 2021-07-01
Fix a bug in slice operator

Sand3r-:
The len() method was used on a tensor's shpae, which is no longer legal in TF2.


chinhuang007:
tf.shape(x) should work as expected, don't see the need to change to x.get_shape()

Not sure where the bug is... please provide the data and code to show the problem.

Sand3r-:
I don't quite recall what the exact error message was, but with TF 2.4 I had an error saying that calling __len__ on tensor shape is ambigious during my conversion attempt. Unfortunately I cannot provide the data due to confidentiality measures.

chinhuang007 (approved):
LGTM!

PR #940, opened 2021-07-05
support sequence type; support opset 14 of identity op

chudegao:

### DIFF
--- a/onnx_tf/handlers/backend/identity.py
+++ b/onnx_tf/handlers/backend/identity.py
@@ -6,11 +6,9 @@ from onnx_tf.handlers.handler import tf_func
 from onnx_tf.handlers.handler import partial_support
 from onnx_tf.handlers.handler import ps_description
 
+
 @onnx_op("Identity")
 @tf_func(tf.identity)
-@partial_support(True)
-@ps_description("Identity with sequence type is not supported " +
-                "in Tensorflow")
 class Identity(BackendHandler):
 
   @classmethod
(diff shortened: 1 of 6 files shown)


onnx_tf/handlers/backend/identity.py
```diff
@@ -20,3 +16,11 @@ def version_1(cls, node, **kwargs):
   @classmethod
   def version_13(cls, node, **kwargs):
     return [cls.make_tensor_from_onnx_node(node, **kwargs)]
+
+  @classmethod
+  def version_14(cls, node, **kwargs):
+    x = kwargs["tensor_dict"][node.inputs[0]]
+    if isinstance(x, (list, tuple)):
+      return [tf.identity_n(x)]
```
chinhuang007, line 24: Please note ONNX sequence allows variable shapes and tf.identity_n supports tensors with same shapes only.

chinhuang007:
In general, an ONNX sequence can have tensors with difference shapes. We should cover such case with code and tests to claim the complete support.
Currently the standard ONNX backend tests for identity with sequences are commented out in test/backend/test_onnx_backend.py, line 148. Once uncommented, we don't pass the standard tests with this patch.

chudegao:
onnx backend test have an issue to support seqence output, I submit a PR to onnx to fix it https://github.com/onnx/onnx/pull/3563. After it's merged I will push new commit in this pr, otherwise test_identity_sequence will still fail.

chinhuang007 (approved):
LGTM!

Issue #951, opened 2021-07-20
ValueError: Dimensions must be equal, but are 32 and 8 for '{{node Add_4}} = Add[T=DT_FLOAT](concat_4/concat, Add_4/y)' with input shapes: [?,51,8,32], [8].


nikhiltiru:
**Describe the bug**

Cannot convert onnx model to Tensorflow.

Converted PyTorch model to onnx.
Checked that onnx model works using onnxruntime
Converting from onnx to Intel OpenVino also works.
But converting to TensorFlow fails.

**To Reproduce**

```
import onnx
import onnx_tf

onnx_path = 'float_model.onnx'
onnx_model = onnx.load(onnx_path)
tf_rep = onnx_tf.backend.prepare(onnx_model)
tf_rep.export_graph("model.tf")
```

**ONNX model file**

https://drive.google.com/file/d/1zax7qCOxF0dCQ2VYOwyF1kG_eCpn6sk7/view?usp=sharing

**PyTorch model file**

https://drive.google.com/file/d/1UIKKECuanMlb0ilxvPNn_MrUNI-wSpyf/view?usp=sharing

**Python, ONNX, ONNX-TF, Tensorflow version**

This section can be obtained by running `get_version.py` from util folder.
 - Python version: 3.8.2 [GCC 8.3.0]
 - ONNX version: 1.9.0
 - ONNX-TF version: 1.8.0
 - Tensorflow version: 2.5.0

**Additional context**

Stack Trace
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/tmp/ipykernel_14490/269763652.py in <module>
      5 onnx_model = onnx.load(onnx_path)
      6 tf_rep = onnx_tf.backend.prepare(onnx_model)
----> 7 tf_rep.export_graph("model.tf")

~/.local/lib/python3.8/site-packages/onnx_tf/backend_rep.py in export_graph(self, path)
    113         self.tf_module,
    114         path,
--> 115         signatures=self.tf_module.__call__.get_concrete_function(
    116             **self.signatures))

~/.local/lib/python3.8/site-packages/tensorflow/python/eager/def_function.py in get_concrete_function(self, *args, **kwargs)
   1365       ValueError: if this object has not yet been called on concrete values.
   1366     """
-> 1367     concrete = self._get_concrete_function_garbage_collected(*args, **kwargs)
   1368     concrete._garbage_collector.release()  # pylint: disable=protected-access
   1369     return concrete

[...]

PR #954, opened 2021-07-28
Fix issue 951

chudegao:
In issue https://github.com/onnx/onnx-tensorflow/issues/951, when batch_size is dynamic value[None, 1, 156, 26] and conv layer's group attr is not 1. It will set depthwise=True without checking group value and x_shape[1] . However, if batch_size is fix value, it will reset depthwise=(group == x_shape[1]). 

    depthwise = (x_rank == 4 and len(weight_shape) == 4 and group != 1 and
                 not transpose and not (None in weight_shape))
    if depthwise and isinstance(x_shape, np.ndarray):
      depthwise = bool(group == x_shape[1])

To fix it, check group value and x_shape[1] always.


onnx_tf/handlers/backend/conv_mixin.py
```diff
@@ -97,8 +95,8 @@ def conv(cls, node, input_dict, transpose=False):
     # Is this convolution depthwise we can support?
     depthwise = (x_rank == 4 and len(weight_shape) == 4 and group != 1 and
                  not transpose and not (None in weight_shape))
-    if depthwise and isinstance(x_shape, np.ndarray):
-      depthwise = bool(group == x_shape[1])
+    if depthwise and (x_rank > 1 and x.get_shape().as_list()[1] != None):
```
chinhuang007, line 98: It might not be necessary to check x_rank > 1 since depthwise is true, meaning x_rank == 4

chudegao: Yes. Thanks. Remove x_rank > 1.

chinhuang007 (approved):
LGTM!

Issue #953, opened 2021-07-24
GatherAndScatterMixin: TypeError: slice indices must be integers or None or have an __index__ method


mmeendez8:
**Describe the bug**

I am trying to convert an ONNX model exported from [mmdetection](https://github.com/open-mmlab/mmdetection) framework but I am getting this error:

```
2021-07-24 13:24:50.300861: I tensorflow/stream_executor/platform/default/dso_loader.cc:53] Successfully opened dynamic library libcudart.so.11.0
2021-07-24 13:24:52.419083: I tensorflow/stream_executor/platform/default/dso_loader.cc:53] Successfully opened dynamic library libcuda.so.1
2021-07-24 13:24:52.463692: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1733] Found device 0 with properties: 
pciBusID: 0000:18:00.0 name: GeForce RTX 3090 computeCapability: 8.6
coreClock: 1.695GHz coreCount: 82 deviceMemorySize: 23.70GiB deviceMemoryBandwidth: 871.81GiB/s
2021-07-24 13:24:52.464971: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1733] Found device 1 with properties: 
pciBusID: 0000:86:00.0 name: GeForce RTX 3090 computeCapability: 8.6
coreClock: 1.695GHz coreCount: 82 deviceMemorySize: 23.70GiB deviceMemoryBandwidth: 871.81GiB/s
2021-07-24 13:24:52.465850: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1733] Found device 2 with properties: 
pciBusID: 0000:3b:00.0 name: GeForce RTX 2080 Ti computeCapability: 7.5
coreClock: 1.635GHz coreCount: 68 deviceMemorySize: 10.76GiB deviceMemoryBandwidth: 573.69GiB/s
2021-07-24 13:24:52.466728: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1733] Found device 3 with properties: 
pciBusID: 0000:af:00.0 name: GeForce RTX 2080 Ti computeCapability: 7.5
coreClock: 1.635GHz coreCount: 68 deviceMemorySize: 10.76GiB deviceMemoryBandwidth: 573.69GiB/s
2021-07-24 13:24:52.466754: I tensorflow/stream_executor/platform/default/dso_loader.cc:53] Successfully opened dynamic library libcudart.so.11.0
[...]

PR #956, opened 2021-07-30
Fix issue 953

chudegao:
Fix issue https://github.com/onnx/onnx-tensorflow/issues/953.

From the error below, indices_shape[-1] is not a right value as slice.
    /Users/degaochu/git/onnx-tensorflow/onnx_tf/handlers/backend/gather_and_scatter_mixin.py:76 process_neg_idx  *
        max_i = tf.cast(data_shape[:indices_shape[-1]], indices.dtype)
    TypeError: slice indices must be integers or None or have an __index__ method

debug info:
data: Tensor("onnx_tf_prefix_ConstantOfShape_2607:0", shape=(4741, 1), dtype=float32)
indices:  Tensor("onnx_tf_prefix_Concat_2639:0", shape=(None, None, 2), dtype=int64)
indices_shape: Tensor("Shape_648:0", shape=(3,), dtype=int64) 

From the debug info, indices is [None, None, 2], indices_shape = tf_shape(indices). When data_shape is ndarray, indices_shape[-1] is a tensor which can not be used as slice index of ndarray.  

To fix it, use indices.get_shape()[-1] when data_shape is ndarray .


chinhuang007 (approved):
LGTM!

PR #958, opened 2021-08-03
Partially support BatchNormalization opset 14

chinhuang007:
Partially support BatchNormalization opset 14, where the new
training_mode attribute is not 1


chudegao (approved):
LGTM

PR #960, opened 2021-08-13
made a branch that copies from tag 1.7.0-tf-1.15

sherylZhaoCode:


chinhuang007:
Hi, @sherylZhaoCode, this PR is attempting to make changes to the master branch with older versions, instead of creating a new branch. I am not sure that is the true intent based on the PR title...

If the goal is to find a version of onnx-tf that works with onnx 1.7 and tf 1.15, please checkout the tf-1x branch, as the supported versions for that branch can be seen here, https://github.com/onnx/onnx-tensorflow/blob/tf-1.x/Versioning.md.

chinhuang007:
We can't change the master branch to support older versions, therefore close the PR. Please use the tf-1x branch instead for Tensorflow 1.x related issues and PRs. Thanks.

Issue #965, opened 2021-08-22
backend.prepare input_tensor_dict argument not useful because of new graph creation


anishathalye:
It would be nice if the `input_tensor_dict` keyword argument to `backend.prepare()` was actually useful, but right now, because the code [creates a new graph](https://github.com/onnx/onnx-tensorflow/blob/72c81448bb17ca4adcd6c4f55e8e84ee0aef445a/onnx_tf/backend.py#L131), it's not usable. (The caller can't create a tensor that's in the same graph that the code is going to use, because that graph isn't created until later.)

If we try to pass in a placeholder created outside/before the call to prepare(), we get the following kind of error:

```
ValueError: Tensor("split:0", shape=(3, 3, 3, 16), dtype=float32) must be from the same graph as Tensor("transpose:0", shape=(1, 360, 360, 3), dtype=float32) (graphs are <tensorflow.python.framework.ops.Graph object at 0x7f129c388748> and <tensorflow.python.framework.ops.Graph object at 0x7f129cc90048>).
```

I'm not familiar with the implementation details of onnx-tensorflow. Would it be reasonable to modify `prepare()` to take a `graph=` keyword argument, and then use the supplied graph instead of always creating a new graph?

chudegao:
@anishathalye can you show some example code how to reproduce this issue and what's your use case.   Currently, graph is only used for special case(training).   Maybe you can try create a new tensor as value of input_tensor_dict to avoid graph conflict.

anishathalye:
Yes, I was using this for training-like purposes; I needed to compute derivatives through the model for the purpose of producing [adversarial examples for NeuralHash](https://github.com/anishathalye/neural-hash-collider). I'm not using this fancier functionality in the public repo, but here's essentially what I was trying to do that led to the error mentioned above:

```python
a = tf.placeholder(...)
b = f(a)  # calls some tensorflow functions, creates some more graph nodes...
model = backend.prepare(input_tensor_dict={image: b}, training_mode=True)
...
g, = tf.gradients(model.tensor_dict['output'], a)
```

[...]

PR #968, opened 2021-08-29
support use user tensorflow graph when training mode is enabled

chudegao:

Fix issue https://github.com/onnx/onnx-tensorflow/issues/965


chinhuang007 (approved):
LGTM!

PR #977, opened 2021-09-21
Add support for Shape opset 15

chinhuang007:
Add support for Shape opset 15 where two optional attributes, start
and end, are introduced.


chudegao (approved):
LGTM!

PR #991, opened 2021-11-24
support bernoulli opset 15

chudegao:

This operator is non-deterministic and may not produce the same values so we still need to exlucde the test case.

### DIFF
--- a/onnx_tf/handlers/backend/bernoulli.py
+++ b/onnx_tf/handlers/backend/bernoulli.py
@@ -3,7 +3,6 @@ from tensorflow_probability import distributions as tfd
 
 from onnx_tf.handlers.backend_handler import BackendHandler
 from onnx_tf.handlers.handler import onnx_op
-from onnx_tf.common import data_type
 from onnx_tf.handlers.handler import partial_support
 from onnx_tf.handlers.handler import ps_description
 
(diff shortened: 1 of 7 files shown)


onnx_tf/handlers/backend/bernoulli.py
```diff
@@ -0,0 +1,31 @@
...
+@ps_description(
+    "Bernoulli with float type seed will be converted to int type seed")
+class Bernoulli(BackendHandler):
+
+  @classmethod
+  def _common(cls, node, **kwargs):
+    x = kwargs["tensor_dict"][node.inputs[0]]
+    dtype = x.dtype
+    if 'dtype' in node.attrs:
+      dtype = node.attrs.get('dtype')
```
chinhuang007, line 21: Could be combined to one line, dtype = node.attrs.get("dtype", x.dtype)

chinhuang007:
Everything looks good. Can you re-apply cla?

onnx_tf/handlers/backend/bernoulli.py
chudegao: Thanks, updated.

chudegao:
> Everything looks good. Can you re-apply cla?

I'm not sure how to handle this. Can you update status manually. I hit this issues before. Sometimes, it will pass the check after serveral hours. Sometimes it keeps waiting. 

图片

chinhuang007 (approved):
LGTM

PR #996, opened 2021-12-24
update Optional operator to return right type value for empty value

chudegao:

Return None for tensor type and return [] for sequence type.

https://github.com/onnx/onnx/blob/master/docs/Operators.md#Optional
Optional
Constructs an optional-type value containing either an empty optional of a certain type specified by the attribute

Type Constraints
V : tensor(uint8), tensor(uint16), tensor(uint32), tensor(uint64), tensor(int8), tensor(int16), tensor(int32), tensor(int64), tensor(float16), tensor(float), tensor(double), tensor(string), tensor(bool), tensor(complex64), tensor(complex128), seq(tensor(uint8)), seq(tensor(uint16)), seq(tensor(uint32)), seq(tensor(uint64)), seq(tensor(int8)), seq(tensor(int16)), seq(tensor(int32)), seq(tensor(int64)), seq(tensor(float16)), seq(tensor(float)), seq(tensor(double)), seq(tensor(string)), seq(tensor(bool)), seq(tensor(complex64)), seq(tensor(complex128))
    Constrains input type to all tensor and sequence types.

### DIFF
--- a/onnx_tf/handlers/backend/optional.py
+++ b/onnx_tf/handlers/backend/optional.py
@@ -10,4 +10,8 @@ class Optional(BackendHandler):
     if len(node.inputs) > 0:
         return [kwargs["tensor_dict"][node.inputs[0]]]
     else:
-        return [None]
+        if node.attrs.get('type').optional_type.elem_type.HasField('tensor_type'):
+            return [None]
+        else:
+            return [[]]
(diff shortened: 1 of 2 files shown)


chinhuang007 (approved):
LGTM!

PR #1001, opened 2022-01-13
support batch normalization opset 15

chudegao:
https://github.com/onnx/onnx/pull/3545

Description

    Add additional type constraints (T1) for scale and bias of BatchNormalization

Motivation and Context

    Current type binding constrains input X and scale/bias of the same type T, while in PyTorch there are cases where X is of float16 and scale/bias of float, for example. Such cases should be handled.


chinhuang007 (approved):
LGTM!

PR #1018, opened 2022-04-01
Fix input rank unknown issue for split

chinhuang007:
Fix input rank unknown issue for split by checking the input
rank and skipping the validation when rank is unknown.


chudegao (approved):
LGTM!

